PR #1, opened 2019-09-19
feat: Add module pureconfig with tests and basic documentation

jakubjanecek:
feat: Add module example to show how to use the code

build: Tweak the build so that it all works together well

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,19 +7,47 @@ ThisBuild / scmInfo := Some(
   ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
 )
 
-ThisBuild / scalaVersion := "2.13.1"
+ThisBuild / scalaVersion := "2.13.0"
 ThisBuild / scalacOptions := ScalacOptions.default
+
+ThisBuild / turbo := true
 
 lazy val commonSettings = Seq(
   libraryDependencies ++= Seq(
-    compilerPlugin("org.typelevel" %% "kind-projector" % "0.10.3")
+    compilerPlugin("org.typelevel" %% "kind-projector" % "0.10.3"),
+    "org.typelevel" %% "cats-effect" % "2.0.0",
+    "org.scalatest" %% "scalatest" % "3.0.8" % Test
   ),
   Test / publishArtifact := false
 )
 
 lazy val root = (project in file("."))
+  .aggregate(example, pureconfig)
   .settings(
-    commonSettings,
     name := "scala-server-toolkit",
     publish / skip := true
+  )
+
+lazy val example = project
+  .dependsOn(pureconfig)
+  .enablePlugins(MdocPlugin)
+  .settings(
+    commonSettings,
+    name := "scala-server-toolkit-example",
+    publish / skip := true,
+    run / fork := true,
+    Global / cancelable := true,
+    mdocIn := baseDirectory.value / "src" / "main" / "mdoc",
+    mdocOut := baseDirectory.value / ".." / "docs",
+    libraryDependencies ++= Seq(
+      "dev.zio" %% "zio" % "1.0.0-RC12-1",
+      "dev.zio" %% "zio-interop-cats" % "2.0.0.0-RC3"
+    )
+  )
+
+lazy val pureconfig = project
+  .settings(
+    commonSettings,
+    name := "scala-server-toolkit-pureconfig",
+    libraryDependencies += "com.github.pureconfig" %% "pureconfig" % "0.12.0"
   )
--- a/build.sbt
+++ b/build.sbt
@@ -14,9 +14,9 @@ ThisBuild / turbo := true
 
 lazy val commonSettings = Seq(
   libraryDependencies ++= Seq(
-    compilerPlugin("org.typelevel" %% "kind-projector" % "0.10.3"),
-    "org.typelevel" %% "cats-effect" % "2.0.0",
-    "org.scalatest" %% "scalatest" % "3.0.8" % Test
+    compilerPlugin(Dependencies.kindProjector),
+    Dependencies.catsEffect,
+    Dependencies.scalaTest
   ),
   Test / publishArtifact := false
 )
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -7,19 +7,47 @@ ThisBuild / scmInfo := Some(
   ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
 )
 
-ThisBuild / scalaVersion := "2.13.1"
+ThisBuild / scalaVersion := "2.13.0"
```
hanny24: why not 2.13.1?

pureconfig/src/main/scala/com/avast/server/toolkit/pureconfig/PureConfigModule.scala
```diff
@@ -0,0 +1,42 @@
...
+  /** Loads the case class `A` using Lightbend Config's standard behavior. */
+  def make[F[_]: Sync, A: ConfigReader]: F[Either[NonEmptyList[String], A]] = make(ConfigSource.default)
+
+  /** Loads the case class `A` using provided [[pureconfig.ConfigSource]]. */
+  def make[F[_]: Sync, A: ConfigReader](source: ConfigSource): F[Either[NonEmptyList[String], A]] = {
+    Sync[F].delay(source.load[A].leftMap(convertFailures))
+  }
+
+  /** Loads the case class `A` using Lightbend Config's standard behavior or throws an exception. */
+  def makeOrThrow[F[_]: Sync, A: ConfigReader: ClassTag]: F[A] = makeOrThrow(ConfigSource.default)
```
hanny24: I would prefer to call it `makeOrRaise` such that it's clear that the exception is wrapped in `F[_]`

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.avast.server.toolkit.example.config.Configuration
+import com.avast.server.toolkit.pureconfig.PureConfigModule
+import zio.interop.catz._
+import zio.{Task, ZIO}
+
+object Main extends CatsApp {
+
+  def program: Resource[Task, Unit] = {
+    for {
+      configuration <- Resource.liftF(PureConfigModule.makeOrThrow[Task, Configuration])
```
hanny24: I don't think that raising exception is the the best way to solve invalid configuration. I understand it's convenient, but the stack trace is completely useless.

build.sbt
jakubjanecek: Because `2.13.1` does not work with `mdoc`.

build.sbt
```diff
@@ -7,19 +7,47 @@ ThisBuild / scmInfo := Some(
...
+      "dev.zio" %% "zio" % "1.0.0-RC12-1",
+      "dev.zio" %% "zio-interop-cats" % "2.0.0.0-RC3"
+    )
+  )
+
+lazy val pureconfig = project
+  .settings(
+    commonSettings,
+    name := "scala-server-toolkit-pureconfig",
+    libraryDependencies += "com.github.pureconfig" %% "pureconfig" % "0.12.0"
```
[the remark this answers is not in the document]
jakubjanecek: Since most of the submodules will usually depend only on single dependency (which will not be shared with any other module) I don't see the benefit - this way I see directly what depenedency and in which version is used. So I'd like to keep it that way. OK?

pureconfig/src/main/scala/com/avast/server/toolkit/pureconfig/PureConfigModule.scala
jakubjanecek: Alright, sounds good to me.

jakubjanecek:
@augi This module is very simple so the benefit is not substantial. It gives you 3 things: wrapping in `Sync`, improved error message and conventional (according to this project) way of doing things. Of course you can use `pureconfig` directly, nothing against that, but you will have to write `Sync[Task].delay` repeatedly and you might want to wrap this functionality into some class which is basically what this module is doing.

I will write some documentation covering the design of this project but not as part of this PR. Alright?

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
jakubjanecek: I understand your view but I think this approach is pragmatic. Usually loading of the configuration will be the first thing in your application and if that fails you definitely do not want to continue - it's fatal error. This is why I think throwing an exception is just fine, not because of the stack trace which is as you say useless, but because it stops your application immediately and you do not have to wrap the whole initialization into `EitherT` or find other solution.

build.sbt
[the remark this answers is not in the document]
hanny24: I don't think it's true. Non trivial amount of deps will at least depend on cats-effect.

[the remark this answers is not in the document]
jakubjanecek: Yes, `cats-effect` is the only exception and that's why it's part of `commonSettings` shared by all the modules (and defined in single place).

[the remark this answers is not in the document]
hanny24: I still thing it should be extracted into separate file as it is more manageable. I won't block it right now, but mark my words that we will move it to separate file eventually.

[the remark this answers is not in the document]
jakubjanecek: Alright, I did it ;)

docs/pureconfig.md
```diff
@@ -0,0 +1,27 @@
+# Module PureConfig
+
+![GitHub tag (latest SemVer)](https://img.shields.io/github/v/tag/avast/scala-server-toolkit?label=version&sort=semver)
+
+`libraryDependencies += "com.avast.server.toolkit" %% "scala-server-toolkit-pureconfig" % "<VERSION>"`
+
+This module allows you to load your application's configuration file according to a case class provided to it. It uses
+[PureConfig](https://pureconfig.github.io) library to do so which uses [Lightbend Config](https://github.com/lightbend/config) which means
+that your application's configuration will be in [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) format.
+
```
augi: Could you please write the rationale why this module is useful here? If I understand it correctly, `ConfigSource.default.load[A]` is an operation with side-effect and so it should be wrapped in an `F`. And this is what this module is actually fixing, right?

jakubjanecek: OK

PR #8, opened 2019-09-22
docs: Add section about project's rationale

jakubjanecek:


docs/pureconfig.md
```diff
@@ -20,9 +20,9 @@ import zio.Task
 final case class ServerConfiguration(listenAddress: String, listenPort: Int)
 
 implicit val serverConfigurationReader: ConfigReader[ServerConfiguration] = deriveReader
-// serverConfigurationReader: ConfigReader[ServerConfiguration] = pureconfig.generic.DerivedConfigReader1$$anon$3@7d4e424e
+// serverConfigurationReader: ConfigReader[ServerConfiguration] = pureconfig.generic.DerivedConfigReader1$$anon$3@3c04ddda
```
tomasherman: really? :)

example/src/main/mdoc/rationale.md
```diff
@@ -0,0 +1,42 @@
+# Rationale
```
tomasherman: why is this here twice?

docs/pureconfig.md
jakubjanecek: That's generated document with output from console. I'll probably redo this somehow but it is not important for now and it serves as some documentation...

example/src/main/mdoc/rationale.md
jakubjanecek: This document is the source, the other one is generated out of this by mdoc. This document does not contain any code so it could be here only once but I like to keep it consistent and have all the sources at single place.

PR #14, opened 2019-09-24
badge-links

sideeffffect:


README.md
```diff
@@ -1,7 +1,7 @@
 # Scala Server Toolkit
 
 [![Build Status](https://travis-ci.org/avast/scala-server-toolkit.svg?branch=master)](https://travis-ci.org/avast/scala-server-toolkit)
-![Maven Central](https://img.shields.io/maven-central/v/com.avast/scala-server-toolkit-pureconfig_2.13)
+[![Maven Central](https://img.shields.io/maven-central/v/com.avast/scala-server-toolkit-pureconfig_2.13)](https://repo1.maven.org/maven2/com/avast/scala-server-toolkit-pureconfig_2.13/)
```
augi: Isn't it strange that the link from root README goes to `pureconfig` module?

jakubjanecek:
@sideeffffect What is this change? How is it different than what's already in there?

README.md
jakubjanecek: That's temporary because there is no other module released yet :)

jakubjanecek:
> @sideeffffect What is this change? How is it different than what's already in there?

Oh, I see now. The difference is that the badge is actually a link to the repo, right?

PR #16, opened 2019-09-24
Improve my car... ehh... sbt

sideeffffect:
image


.travis.yml
```diff
@@ -10,7 +10,7 @@ before_install:
 stages:
   - name: test
   - name: release
-    if: ((branch = master AND type = push) OR (tag IS present)) AND NOT fork
+    if: ((branch = master) OR (tag IS present)) AND NOT fork
```
jakubjanecek: Can you please explain this?

build.sbt
```diff
@@ -1,27 +1,62 @@
+addCommandAlias("check", "; scalafmtSbtCheck; scalafmtCheckAll; compile:scalafix --check ; test:scalafix --check")
```
jakubjanecek: Please move these commands to the bottom of this file, I don't think that it should be up here.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
+addCommandAlias("fix", "; scalafmtSbt; scalafmtAll; compile:scalafix ; test:scalafix")
+
 ThisBuild / organization := "com.avast"
 ThisBuild / homepage := Some(url("https://github.com/avast/scala-server-toolkit"))
 ThisBuild / description := "Functional programming toolkit for building server applications in Scala."
 ThisBuild / licenses := Seq("MIT" -> url("https://raw.githubusercontent.com/avast/scala-server-toolkit/master/LICENSE"))
 ThisBuild / developers := List(Developer("jakubjanecek", "Jakub Janecek", "janecek@avast.com", url("https://www.avast.com")))
-ThisBuild / scmInfo := Some(
```
jakubjanecek: Why did you remove this? I think it's necessary for the `sbt-ci-release` plugin and if not I think the info is added to the Maven metadata so I'd like to keep it.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
 
 ThisBuild / scalaVersion := "2.13.0"
-ThisBuild / scalacOptions := ScalacOptions.default
 
 ThisBuild / turbo := true
 
 lazy val commonSettings = Seq(
   libraryDependencies ++= Seq(
     compilerPlugin(Dependencies.kindProjector),
+    compilerPlugin(Dependencies.betterMonadicFor),
```
jakubjanecek: I don't want any non-standard syntax like this in the project, please remove this plugin.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
+    compilerPlugin(Dependencies.silencer),
+    compilerPlugin(scalafixSemanticdb), // for scalafix
+    Dependencies.silencerLib,
     Dependencies.catsEffect,
-    Dependencies.scalaTest
+    Dependencies.scalaTest % Test
```
jakubjanecek: I don't know which version you base this on but in newer version of `master` ScalaTest is already marked as `Test` scope so this shouldn't be needed.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
+  ),
+  Compile / compile / wartremoverErrors ++= Warts.all filterNot Set(
```
jakubjanecek: I think this should be refactored to some other file same way compiler options are refactored.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
+  ThisBuild / scalafixDependencies ++= Seq(
+    Dependencies.scalazzi // https://github.com/scalaz/scalazzi
+  ),
+  scalacOptions ++= Seq(
+    "-Yrangepos", // for scalafix. required by SemanticDB compiler plugin
+    "-Ywarn-unused", // for scalafix. not present in sbt-tpolecat for 2.13
+    "-P:silencer:checkUnused"
+  ),
+  scalacOptions --= {
+    if (!sys.env.contains("TRAVIS"))
```
jakubjanecek: Why? I'd like to keep the build as clean as possible from such ifs.

build.sbt
```diff
@@ -1,27 +1,62 @@
...
+    Wart.MutableDataStructures,
+    Wart.OptionPartial,
+    Wart.AsInstanceOf,
+    Wart.EitherProjectionPartial
+  )
 )
 
-lazy val root = (project in file("."))
+lazy val root = project
+  .in(file("."))
```
jakubjanecek: I already did this in `master` so you'll have to rebase/merge.

build.sbt
```diff
@@ -31,8 +66,8 @@ lazy val root = (project in file("."))
 lazy val example = project
   .dependsOn(pureconfig)
   .enablePlugins(MdocPlugin)
+  .settings(commonSettings)
```
jakubjanecek: This change seems petty to me ;)

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
```diff
@@ -3,17 +3,19 @@ package com.avast.server.toolkit.example
...
 object Main extends CatsApp {
 
   def program: Resource[Task, Unit] = {
     for {
-      configuration <- Resource.liftF(PureConfigModule.makeOrRaise[Task, Configuration])
+      _ <- Resource.liftF(PureConfigModule.makeOrRaise[Task, Configuration])
     } yield ()
   }
 
+  @silent("dead code")
```
jakubjanecek: Which peace of this code is dead? It shouldn't or am I wrong?

project/Dependencies.scala
```diff
@@ -2,11 +2,26 @@ import sbt._
 
 object Dependencies {
 
-  val catsEffect = "org.typelevel" %% "cats-effect" % "2.0.0"
-  val kindProjector = "org.typelevel" %% "kind-projector" % "0.10.3"
-  val pureConfig = "com.github.pureconfig" %% "pureconfig" % "0.12.0"
-  val scalaTest = "org.scalatest" %% "scalatest" % "3.0.8" % Test
-  val zio = "dev.zio" %% "zio" % "1.0.0-RC13"
-  val zioInteropCats = "dev.zio" %% "zio-interop-cats" % "2.0.0.0-RC4"
+  object Versions {
```
jakubjanecek: I don't like this. It seems unnecessary to me because you are duplicating the name of the library 3 times in this file. I was even against having the `Dependencies` file because 99% of the dependencies are defined only once in the whole project so the extraction is unenecessary but in the end I said ok let's have it refactored into a file but extracting also versions seems just crazy to me. Please don't do this. And please keep the dependencies sorted. Also, in master ScalaTest is already refactored into "test" dependencies.

.travis.yml
sideeffffect: I'm hoping that this will publish from master even when not on a tag, i.e. a snapshot

build.sbt
sideeffffect: `sbt-ci-release` seems to be already adding this

https://github.com/olafurpg/sbt-ci-release/blob/master/plugin/src/main/scala/com/geirsson/CiReleasePlugin.scala#L79

build.sbt
sideeffffect: I've had so far only good experience with this plugin, but if you think it won't make your life better ;)

build.sbt
sideeffffect: I've always (mostly?) seen people declare `% Test` explicitly, so I'm just following a convention

build.sbt
sideeffffect: ok, that is probably a good idea

build.sbt
sideeffffect: I think it's worth failing warnings only on CI, because locally `fatal warnings` doesn't work well with `scalafix`, eg https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html

build.sbt
sideeffffect: yep

build.sbt
sideeffffect: yeah, I did it mindlessly when trying to concentrate :D

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
sideeffffect: false positive :(

project/Dependencies.scala
sideeffffect: I think this is the most valuable part of pushing "dependencies" away into a common place -- very often you want to use the same version for what is logically one library, but technically is multiple jars (like this project, or cats/cats-mtl, and so on)

doing it this style makes you declare that version only once

sideeffffect:
@jakubjanecek build is green, ready to merge from my POV. we can talk about details in person tomorrow if you want

example/lock.sbt
```diff
@@ -0,0 +1,113 @@
+// DON'T EDIT THIS FILE.
+// This file is auto generated by sbt-lock 0.6.1.
+// https://github.com/tkawachi/sbt-lock/
+dependencyOverrides ++= {
```
tomasherman: i don't like this at all - what problem does it solve?

jvm-execution/src/main/scala/com/avast/server/toolkit/execution/ConfigurableThreadFactory.scala
```diff
@@ -35,6 +35,7 @@ object ConfigurableThreadFactory {
   /**
     * @param nameFormat Formatted with long number, e.g. my-thread-%02d
     */
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
```
tomasherman: why are we calling DefaultArguments wart? if it's a wart, why are we using it?

build.sbt
```diff
@@ -3,22 +3,16 @@ ThisBuild / homepage := Some(url("https://github.com/avast/scala-server-toolkit"
 ThisBuild / description := "Functional programming toolkit for building server applications in Scala."
 ThisBuild / licenses := Seq("MIT" -> url("https://raw.githubusercontent.com/avast/scala-server-toolkit/master/LICENSE"))
 ThisBuild / developers := List(Developer("jakubjanecek", "Jakub Janecek", "janecek@avast.com", url("https://www.avast.com")))
-ThisBuild / scmInfo := Some(
-  ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
-)
 
 ThisBuild / scalaVersion := "2.13.0"
```
tomasherman: btw are we not building this for 2.12? Why?

build.sbt
jakubjanecek: I wouldn't call it a convention, I haven't seen it that much. Anyway, the difference between our approaches is very small. I am not against it, just I think you should now do whatever is in `master` and we can do another PR to update this style, ok?

build.sbt
jakubjanecek: And this is exactly what I was talking about! The linters are not mature in my opinion because they have these limitations I don't like. I want my build to fail on warnings locally the same way it fails in CI. It's not acceptable for me to have it behave differently. I'd rather get rid of scalafix than this, sorry.

example/lock.sbt
jakubjanecek: I also think right now we are not in a state where we would need this. We can always add it but I'd like to remove it for now too.

jvm-execution/src/main/scala/com/avast/server/toolkit/execution/ConfigurableThreadFactory.scala
jakubjanecek: @sideeffffect Can you explain the wart? I am using default arguments on purpose and having to suppress it everywhere in my code seems very strange, I'd rather remove the wart.

build.sbt
jakubjanecek: Well, idea that it could only be built against `2.13` maybe :) But I guess it won't fly... we can update that later before first stable release.

.scalafix.conf
```diff
@@ -0,0 +1,35 @@
...
+]
+
+ExplicitResultTypes {
+  unsafeShortenNames = true
+}
+
+Disable.ifSynthetic = [
+  "java/io/Serializable"
+  "scala/Any"
+  ## "scala/Product" # triggers even on non-synthetic
```
jakubjanecek: This file should only contain rules we want to have and almost sure work well for us. Please remove the commented parts and maybe put some comments why something is enabled or a link to some explanation.

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
```diff
@@ -25,6 +27,7 @@ object Main extends CatsApp {
     } yield ()
   }
 
+  @silent("dead code") // false positive
```
jakubjanecek: And again... we are putting something to our build to protect us but in the end it just bothers us and we have to put comments like this in the code. More and more I am against this overall, sorry.

.travis.yml
jakubjanecek: I'll have to check this, my understanding is very limited however I think snapshots are already published.

jakubjanecek:
@hanny24 @augi Inviting you to the discussion because there seems to be quite big disagreement on how build should look. I am interested in your opinion on the setup of linters (scalafix, wart remover) and use of extra compiler plugins (silencer) My view is that for now it seems to just complicate the build and the code (forcing us to suppress warnings where we shouldn't have) and it does not give us much now - it would protect us probably from some trouble in the future but now I think I am a better linter myself :)

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
```diff
@@ -7,14 +7,16 @@ import com.avast.server.toolkit.example.config.Configuration
 import com.avast.server.toolkit.execution.ExecutorModule
 import com.avast.server.toolkit.pureconfig.PureConfigModule
 import com.avast.server.toolkit.system.console.{Console, ConsoleModule}
+import com.github.ghik.silencer.silent
 import zio.interop.catz._
 import zio.{Task, ZIO}
 
 object Main extends CatsApp {
 
+  @SuppressWarnings(Array("org.wartremover.warts.StringPlusAny")) // false positive
```
augi: I would prefer to disable this wart completely as it can produce false positive.

jvm-execution/src/main/scala/com/avast/server/toolkit/execution/ThreadPoolExecutorConfig.scala
```diff
@@ -4,6 +4,7 @@ import java.util.concurrent.TimeUnit
 
 import scala.concurrent.duration.{Duration, FiniteDuration}
 
+@SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
```
augi: We definitely want to remove this wart.

project/Dependencies.scala
```diff
@@ -2,16 +2,31 @@ import sbt._
...
 
-  val catsEffect = "org.typelevel" %% "cats-effect" % "2.0.0"
-  val kindProjector = "org.typelevel" %% "kind-projector" % "0.10.3"
-  val pureConfig = "com.github.pureconfig" %% "pureconfig" % "0.12.0"
-  val slf4jApi = "org.slf4j" % "slf4j-api" % "1.7.28"
-  val zio = "dev.zio" %% "zio" % "1.0.0-RC13"
-  val zioInteropCats = "dev.zio" %% "zio-interop-cats" % "2.0.0.0-RC4"
+  object Versions {
+    val catsEffect = "2.0.0"
+    val kindProjector = "0.10.3"
```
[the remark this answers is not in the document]
tomasherman: it is - see here: https://github.com/avast/datadog4s/pull/43/files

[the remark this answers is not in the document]
jakubjanecek: Yes, this needs to work with Steward, good point.

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
jakubjanecek: @sideeffffect Are you kidding me? I haven't noticed this but tell me where in this piece of code I am doing anything with strings and `+`? That's ridiculous, the linters do not work well at all.

build.sbt
hanny24: I would also vote to put `% Test` explicitly.

project/Dependencies.scala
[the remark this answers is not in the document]
jakubjanecek: No, the point is that the versions are now refactored into separate `Version` object and I would be afraid it does not work that way...

build.sbt
hanny24: I think what you really want it linters to only warn locally (because it's very annoying if you have some work-in-progress and it fails because of unused import), but also have an easy way how to turn them fatal.

project/Dependencies.scala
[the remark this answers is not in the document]
tomasherman: yeah that might be a problem

jakubjanecek:
@sideeffffect I think @hanny24 is right about the fact that by redoing the whole build it triggered a lot of alarms. I know you have put a lot of work into this and you mean it well however as you can see different people have different opinions so as a whole I don't think we can merge this. I suggest we close this PR and try to add some of the things one by one so that we can digest it slowly. The things I think can be done are the following:
* Adding scalafix but with less rules (the ones that are not controversial and work well across the codebase.
* Marking test dependencies at the place of use. However please keep the versions inlined and do not refactor them to a special object (there's really no need for that, at least now).
* Removing the SCM info.
* Maybe adding some warts but definitely not all of them like now.

And that's it. Let's start with this and we'll update in the future according to our needs. What do you say?

.scalafix.conf
sideeffffect: this is the comment why it's disabled
```
# triggers even on non-synthetic
```
the file is originally from [`scalaz/scalazzi`](https://github.com/scalaz/scalazzi/blob/master/scalafix.conf)
but I wouldn't suggest fullblown scalazzi now, this is just a preparation, so that we can make the step to enforce in the future

example/lock.sbt
sideeffffect: let's not worry about `lock` right now. removed.

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
sideeffffect: as the comment suggests, it's a false possitive
I think we can tolerate one occurence, because it could save us a lot of trouble

jvm-execution/src/main/scala/com/avast/server/toolkit/execution/ConfigurableThreadFactory.scala
sideeffffect: default arguments are frowned upon in the FP community, I personally avoid them too
see https://blog.ssanj.net/posts/2019-05-01-why-are-default-parameter-values-considered-bad-in-scala.html
but this wart applies also to constructors too, which is a big part of PureConfig functionality
so I'm disabling the wart

jvm-execution/src/main/scala/com/avast/server/toolkit/execution/ThreadPoolExecutorConfig.scala
sideeffffect: :heavy_check_mark:

project/Dependencies.scala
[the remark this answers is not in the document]
sideeffffect: it does
https://github.com/typelevel/frameless/pull/397/files

.travis.yml
sideeffffect: you're right! https://oss.sonatype.org/content/repositories/snapshots/com/avast/scala-server-toolkit-http4s-blaze-server_2.12/

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
sideeffffect: I think we can tolerate one or two false positives, it would still be a net benefit, because it catches real errors too

project/BuildHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+    ThisBuild / scalafixDependencies ++= Seq(
+      Dependencies.scalazzi // https://github.com/scalaz/scalazzi
+    ),
+    scalacOptions ++= Seq(
+      "-Yrangepos", // for scalafix. required by SemanticDB compiler plugin
+      "-Ywarn-unused", // for scalafix. not present in sbt-tpolecat for 2.13
+      "-P:silencer:checkUnused"
+    ),
+    scalacOptions --= {
+      if (!sys.env.contains("TRAVIS"))
```
jakubjanecek: Please remove this, I don't think it's needed for now.

.scalafix.conf
```diff
@@ -0,0 +1,40 @@
...
+
+  # local type inference + covariant types fires this
```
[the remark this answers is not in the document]
sideeffffect: :heavy_check_mark:

example/src/main/scala/com/avast/server/toolkit/example/Main.scala
```diff
@@ -26,6 +27,7 @@ object Main extends CatsApp {
     } yield ()
   }
 
+  @silent("dead code") // false positive
```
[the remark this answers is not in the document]
sideeffffect: this isn't any linter, like wartremover or scalafix
this is caused by `scalac` itself, by `-Ywarn-dead-code` specifically
if it's only a rare false positive, I wouldn't worry about it ;)

project/BuildHelper.scala
sideeffffect: I think this is a good idea to fail on warnings in CI
locally, you don't want to fail on warnings, because it break `scalafix` workflow unfortunatelly (see the comment in the file below)

build.sbt
```diff
@@ -3,18 +3,13 @@ ThisBuild / homepage := Some(url("https://github.com/avast/scala-server-toolkit"
...
-  ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
-)
 
 ThisBuild / scalaVersion := "2.12.10"
-ThisBuild / scalacOptions := ScalacOptions.default
 
 ThisBuild / turbo := true
 
-lazy val commonSettings = Seq(
+lazy val commonSettings = BuildHelper.settingsCommon ++ Seq(
```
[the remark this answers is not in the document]
sideeffffect: I think this is what we want. `BuildHelper.settingsCommon` contains common settings for Avast Scala projects. `commonSettings` contains common settings specific for `scala-server-toolkit`. Those 2 are not necessarily the same thing.
`BuildHelper.settingsCommon` for example doesn't have any dependency on `cats` or `scalatest`, and it shouldn't.

PR #17, opened 2019-09-25
build: Switch to Scala 2.12 because the ecosystem is not ready yet

jakubjanecek:
There is no stable version of http4s for 2.13.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ ThisBuild / scmInfo := Some(
   ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
 )
 
-ThisBuild / scalaVersion := "2.13.0"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -7,7 +7,7 @@ ThisBuild / scmInfo := Some(
   ScmInfo(url("https://github.com/avast/scala-server-toolkit"), "scm:git:git@github.com:avast/scala-server-toolkit.git")
 )
 
-ThisBuild / scalaVersion := "2.13.0"
+ThisBuild / scalaVersion := "2.12.10"
```
tomasherman: We can start xcompiling, or is there a reason not to?

jakubjanecek: The reason is stated in the description :) There is no stable version of http4s cross-built to both `2.12` and `2.13` so I think we can now be on `2.12` and cross-build to `2.13` once the ecosystem catches up.

PR #18, opened 2019-09-25
feat: Add http4s client and server modules with simple tests

jakubjanecek:
This is not fully finished but I thought I'd put it up for review earlier. Docs and examples are missing, will be added before merging.


http4s-server/src/main/scala/com/avast/server/toolkit/http4s/CorrelationIdMiddleware.scala
```diff
@@ -0,0 +1,44 @@
...
+        for {
+          newCorrelationId <- OptionT.liftF(Sync[F].delay(UUID.randomUUID()))
+          _ <- log(newCorrelationId)
+          response <- routes(request.withHeaders(request.headers.put(Header(CorrelationIdHeaderName.value, newCorrelationId.toString))))
+                       .map(r => r.withHeaders(r.headers.put(Header(CorrelationIdHeaderName.value, newCorrelationId.toString))))
+        } yield response
+    }
+  }
+
+  private def log[F[_]: Sync](newCorrelationId: UUID) = {
```
tomasherman: you definitely don't want to fix correlationId to uuid

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sRouting.scala
```diff
@@ -0,0 +1,14 @@
...
+import cats.Functor
+import org.http4s.syntax.kleisli._
+import org.http4s.{HttpApp, HttpRoutes}
+
+import scala.language.higherKinds
+
+object Http4sRouting {
+
+  /** Makes [[org.http4s.HttpApp]] from [[org.http4s.HttpRoutes]] */
+  def make[F[_]: Functor](routes: HttpRoutes[F]): HttpApp[F] = routes.orNotFound
```
hanny24: Can you accept multiple routes using varargs?

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientModule.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.avast.server.toolkit.ssl.{SslContextConfig, SslContextModule}
+import javax.net.ssl.SSLContext
+import org.http4s.client.Client
+import org.http4s.client.blaze.BlazeClientBuilder
+
+import scala.concurrent.ExecutionContext
+import scala.language.higherKinds
+
+/** Provides single initialized HTTP client and a method to create more of them. */
+class Http4sClientModule[F[_]: ConcurrentEffect](val client: Client[F]) {
```
sideeffffect: `class Http4sClientModule` doesn't add any value. Having `object Http4sClientModule` with `make` is enough.

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientModule.scala
```diff
@@ -0,0 +1,62 @@
...
+  def makeClient(config: Http4sClientConfig, executionContext: ExecutionContext): Resource[F, Client[F]] = {
+    Http4sClientModule.makeClient(config, executionContext)
+  }
+
+}
+
+object Http4sClientModule {
+
+  /** Makes [[com.avast.server.toolkit.http4s.Http4sClientModule]] with initialized HTTP client. */
+  def make[F[_]: ConcurrentEffect](config: Http4sClientConfig, executionContext: ExecutionContext): Resource[F, Http4sClientModule[F]] = {
```
sideeffffect: I would remove this

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientModule.scala
```diff
@@ -0,0 +1,62 @@
...
+    makeClient(config, executionContext).map(new Http4sClientModule(_))
+  }
+
+  /** Makes [[org.http4s.client.Client]] initialized with the given config. */
+  def makeClient[F[_]: ConcurrentEffect: Applicative](config: Http4sClientConfig,
```
sideeffffect: I would rename this to just `make`

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sRouting.scala
```diff
@@ -0,0 +1,14 @@
+package com.avast.server.toolkit.http4s
+
+import cats.Functor
+import org.http4s.syntax.kleisli._
+import org.http4s.{HttpApp, HttpRoutes}
+
+import scala.language.higherKinds
+
+object Http4sRouting {
```
sideeffffect: are you planning to add more functionality into `Http4sRouting`? I'm asking, because just calling `.orNotFound` doesn't seem much.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+  socketOptions: SocketOptions = SocketOptions.Default
+)
+
+object Http4sServerConfig {
+
+  final case class SocketOptions(tcpNoDelay: Boolean = true)
+
+  object SocketOptions {
+
+    val Default = SocketOptions()
```
sideeffffect: I like the idea of having "default" instances of config case classes! let's make them for every.
but let's name them lowercase `default`.

project/Dependencies.scala
```diff
@@ -3,16 +3,15 @@ import sbt._
 object Dependencies {
 
   val catsEffect = "org.typelevel" %% "cats-effect" % "2.0.0"
+  val http4sBlazeClient = "org.http4s" %% "http4s-blaze-client" % "0.20.11"
+  val http4sBlazeServer = "org.http4s" %% "http4s-blaze-server" % "0.20.11"
+  val http4sDsl = "org.http4s" %% "http4s-dsl" % "0.20.11"
```
sideeffffect: this is a good example, where having versions of packages in `Versions` is good ;)

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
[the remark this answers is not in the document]
jakubjanecek: Not really. All the other configuration case classes have it as default constructor arguments too and the reason for that is when something like PureConfig reads it it can use the default value if it is missing in the configuration file. I don't really know why I put the `Default` variable here, probably just so that it looks a bit better in the `Http4sServerConfig` case class. I think I'll remove it.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+import java.util.concurrent.TimeUnit
+
+import com.avast.server.toolkit.http4s.Http4sServerConfig.SocketOptions
+import org.http4s.blaze.channel
+import org.http4s.server.defaults
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+final case class Http4sServerConfig(
+  listenAddress: String = "127.0.0.1",
```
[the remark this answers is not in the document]
jakubjanecek: You are right. Having it simplifies some tests and example codes and is not really harmful but I agree that no default value makes sense generally. I'll remove it.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/CorrelationIdMiddleware.scala
```diff
@@ -0,0 +1,44 @@
...
+  val CorrelationIdHeaderName = CaseInsensitiveString("Correlation-Id")
+
+  def wrap[F[_]: Sync](routes: HttpRoutes[F]): HttpRoutes[F] = Kleisli[OptionT[F, *], Request[F], Response[F]] { request =>
+    request.headers.get(CorrelationIdHeaderName) match {
+      case Some(header) => routes(request).map(r => r.withHeaders(r.headers.put(Header(CorrelationIdHeaderName.value, header.value))))
+      case None =>
+        for {
+          newCorrelationId <- OptionT.liftF(Sync[F].delay(UUID.randomUUID()))
+          _ <- log(newCorrelationId)
+          response <- routes(request.withHeaders(request.headers.put(Header(CorrelationIdHeaderName.value, newCorrelationId.toString))))
```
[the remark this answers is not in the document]
jakubjanecek: I did not know about attributes, you are right.

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientModule.scala
augi: Or rename to `makeBlazeClient`.

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.http4s.headers.{`User-Agent`, AgentComment, AgentProduct}
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+final case class Http4sClientConfig(
+  responseHeaderTimeout: Duration = Duration.Inf,
+  idleTimeout: FiniteDuration = Duration(1, TimeUnit.MINUTES),
+  requestTimeout: FiniteDuration = defaults.RequestTimeout,
+  connectTimeout: FiniteDuration = defaults.ConnectTimeout,
+  userAgent: `User-Agent` = `User-Agent`(AgentProduct("http4s-blaze", Some(BuildInfo.version)), List(AgentComment("Server"))),
```
augi: I would prefer to remove `"-blaze"` substring. E.g. change the _agent_ to `http4s-client`.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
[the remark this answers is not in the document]
augi: I would prefer to change the default value to `0.0.0.0` (it binds to all available addresses).

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+  listenPort: Int = 8080,
```
augi: I would prefer `80` here - it's something that one could expect.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+  nio2Enabled: Boolean = true,
+  webSocketsEnabled: Boolean = false,
+  http2Enabled: Boolean = false,
+  responseHeaderTimeout: FiniteDuration = Duration(defaults.ResponseTimeout.toNanos, TimeUnit.NANOSECONDS),
+  idleTimeout: FiniteDuration = Duration(defaults.IdleTimeout.toNanos, TimeUnit.NANOSECONDS),
+  bufferSize: Int = 64 * 1024,
```
augi: Are these constants the same as in the underlying classes? If so then we could have here `Option[Int]` 🤔

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerModule.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.http4s.server.blaze.BlazeServerBuilder
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.Duration
+import scala.language.higherKinds
+
+object Http4sServerModule {
+
+  /** Makes [[org.http4s.server.Server]] initialized with the given config and [[org.http4s.HttpApp]]. */
+  def make[F[_]: ConcurrentEffect: Timer](config: Http4sServerConfig,
```
augi: I would prefer `makeBlazerServer` name.

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientConfig.scala
sideeffffect: it's better to be explicit here -- we could have different `http4s` clients besides the "default" Blaze one

augi: The problem is the `"default"` :) I think that this configuration should be implementation-agnostic so `http4s-client` is better for me.

Another solution would be renaming the case-class to `BlazeHttp4sClientConfig`.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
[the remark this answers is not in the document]
sideeffffect: I like having a default address in default server config. please don't remove it. `0.0.0.0` is a good default IMHO.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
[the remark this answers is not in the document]
sideeffffect: As I've said in the comment below, I really like the idea of `SocketOptions.default` (and for other config case classes). please don't remove :crying_cat_face:

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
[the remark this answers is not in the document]
tomasherman: i think defaults should be about things that people should only rarely change ... ip address might be ok with 0.0.0.0 - (and even then i would call it something else but default) but for port we should for sure force people to make decision

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerModule.scala
sideeffffect: there are no other `make`like methods, so there's no need to disambiguate

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
tomasherman: i would prefer not having this default here at all ... just make people to make a decision about something as important as this

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientConfig.scala
sideeffffect: `Http4sClientBlazeConfig` perhaps?

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientModule.scala
```diff
@@ -0,0 +1,62 @@
...
+/** Provides single initialized HTTP client and a method to create more of them. */
+class Http4sClientModule[F[_]: ConcurrentEffect](val client: Client[F]) {
+
+  def makeClient(config: Http4sClientConfig, executionContext: ExecutionContext): Resource[F, Client[F]] = {
+    Http4sClientModule.makeClient(config, executionContext)
+  }
+
+}
+
+object Http4sClientModule {
```
sideeffffect: how about renaming this to `Http4sClientBlazeModule`?

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerModule.scala
sideeffffect: to express myself more clearly: I think this module should be dedicate to Blaze server specifically, so the object itself should be called `Http4sServerBlazeModule`. Than the method can be called just `make`

sideeffffect:
@jakubjanecek it seems to me that you intended the modules/config case classes not just for any http4s server/client, but specifically for the Blaze flavours. So why not reflect that in the name of the objects, like `Http4sServerBlazeModule` instead of `Http4sServerModule` (and so on). Then we don't need to have debates about `make` vs `makeBlaze` method naming...

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sRouting.scala
jakubjanecek: Right but the reason why I did it is that you need to know that you need this import `import org.http4s.syntax.kleisli._` which is quite sad and hard to find. Anyway, in the update I will push soon it receives varargs so it gives you a bit more functionality.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
jakubjanecek: Right now I don't see the need for that. Let's add them once we actually need them.

project/Dependencies.scala
jakubjanecek: You are absolutely right but my point is that I can do it now once I actually need it, there is no need to introduce the `Versions` object when all the dependencies are alone. Anyway, I am not going to do it because of Scala Steward, I don't think it could handle, or do you?

http4s-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sClientConfig.scala
jakubjanecek: I think the right name is `Http4sBlazeClientConfig` 😄 But I have good reasons. You want it to start with Http4s so that you can find it easily and the rest just copies the name of the artifact which is `http4s-blaze-client`.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
jakubjanecek: I think guys are right that this value should not have a default so in the next PR update there will be none.

jakubjanecek:
@sideeffffect I agree. After the discussions here I see that I made the mistake of making it generic http4s thing but it's actually only one of the implementations which is `blaze` in this case. I will rename the modules and classes accordingly. Thanks, this is a very good catch because it would be misleading and impossible to get rid of in the future.

jakubjanecek:
I've tried to address all your comments somehow. Please have another look. Thanks.

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Makes [[org.http4s.client.Client]] (Blaze) initialized with the given config. */
+  def make[F[_]: ConcurrentEffect](config: Http4sBlazeClientConfig, executionContext: ExecutionContext): Resource[F, Client[F]] = {
+    makeClient(config, executionContext)
+  }
+
+  private def makeClient[F[_]: ConcurrentEffect: Applicative](config: Http4sBlazeClientConfig,
+                                                              executionContext: ExecutionContext): Resource[F, Client[F]] = {
+    for {
+      maybeSslContext <- Resource.liftF(sslContext(config.sslContext))
```
sideeffffect: maybe the `maybeSslContext` shouldn't be created here, but come as the parameter to the `make` function

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.language.higherKinds
+
+object Http4sBlazeClient {
+
+  /** Makes [[org.http4s.client.Client]] (Blaze) initialized with the given config. */
+  def make[F[_]: ConcurrentEffect](config: Http4sBlazeClientConfig, executionContext: ExecutionContext): Resource[F, Client[F]] = {
+    makeClient(config, executionContext)
+  }
+
+  private def makeClient[F[_]: ConcurrentEffect: Applicative](config: Http4sBlazeClientConfig,
```
sideeffffect: just rename this to `make`

also, no need for `Applicative`, because `ConcurrentEffect` already is

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import com.avast.server.toolkit.ssl.SslContextConfig
+import org.http4s.BuildInfo
+import org.http4s.client.blaze.ParserMode
+import org.http4s.client.defaults
+import org.http4s.headers.{`User-Agent`, AgentComment, AgentProduct}
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+final case class Http4sBlazeClientConfig(
```
sideeffffect: I know I've said it before, but I just want to say again how happy having case classes for configuration makes me! :smile_cat:

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+  responseHeaderTimeout: Duration = Duration.Inf,
+  idleTimeout: FiniteDuration = Duration(1, TimeUnit.MINUTES),
+  requestTimeout: FiniteDuration = defaults.RequestTimeout,
+  connectTimeout: FiniteDuration = defaults.ConnectTimeout,
+  userAgent: `User-Agent` = `User-Agent`(AgentProduct("http4s-client", Some(BuildInfo.version)), List(AgentComment("Server"))),
```
sideeffffect: `"http4s-blaze-client"`?

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeServerConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.avast.server.toolkit.http4s.Http4sBlazeServerConfig.SocketOptions
+import org.http4s.blaze.channel
+import org.http4s.server.defaults
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+final case class Http4sBlazeServerConfig(
+  listenAddress: String,
+  listenPort: Int,
```
sideeffffect: I would welcome the defaults back. I like the "symmetry" between default values and possible config values.

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.http4s.client.Client
+import org.http4s.client.blaze.BlazeClientBuilder
+
+import scala.concurrent.ExecutionContext
+import scala.language.higherKinds
+
+object Http4sBlazeClient {
+
+  /** Makes [[org.http4s.client.Client]] (Blaze) initialized with the given config. */
+  def make[F[_]: ConcurrentEffect](config: Http4sBlazeClientConfig, executionContext: ExecutionContext): Resource[F, Client[F]] = {
```
sideeffffect: I presume the `executionContext` is for callbacks, but it would be great, if it were clear from the name itself ;-)
(applies to similar occurences)

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/CorrelationIdMiddleware.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.{Header, HttpRoutes, Request, Response}
+import org.slf4j.LoggerFactory
+
+import scala.language.higherKinds
+
+/** Provides correlation ID functionality. Either generates new correlation ID for a request or takes the one sent in HTTP header
+  * and puts it to [[org.http4s.Request]] attributes. It is also filled into HTTP response header.
+  */
+class CorrelationIdMiddleware[F[_]: Sync](correlationIdHeaderName: CaseInsensitiveString, generator: () => String) {
```
sideeffffect: can this be a trait?

tomasherman: why would you?

project/Dependencies.scala
sideeffffect: it works! :tada:

https://github.com/typelevel/frameless/pull/397/files

cc @augi

jakubjanecek: OK. But I'll do just the ones that are duplicated, not all of them.

http4s-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sServerConfig.scala
jakubjanecek: Yes, the types and values are taken from `http4s-blaze-server`. If you mean to make it an `Option` in order to let the default values be defined only in `http4s-blaze-server`, that's a good idea but than everything would be `Option` in our case class and that wouldn't be very nice, don't you think?

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
jakubjanecek: It is an option but prefer it this way - it is part of the entire server configuration.

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
jakubjanecek: 💪 👍

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
jakubjanecek: Yes, probably the best option now.

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/CorrelationIdMiddleware.scala
jakubjanecek: It could but I don't think we need to abstract away middleware.

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeServerConfig.scala
jakubjanecek: I don't understand but as I said before I feel like these two values actually do not deserve a default because none is general enough for everybody.

jakubjanecek: However I have added a factory method in case someone needs it - that's a good compromise.

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
jakubjanecek: To be honest I am not sure because the documentation is very limited. I think every `ExecutionContext` not marked specifically (and nowadays probably as `Blocker` from `cats-effect`) should be considered for callbacks. I'll document it.

jakubjanecek:
Everyone I think this is ready for the last round of review and possibly merging.

docs/http4s.md
```diff
@@ -0,0 +1,93 @@
...
+
+[![Maven Central](https://img.shields.io/maven-central/v/com.avast/scala-server-toolkit-http4s-blaze-server_2.12)](https://repo1.maven.org/maven2/com/avast/scala-server-toolkit-http4s-blaze-server_2.12/)
+
+`libraryDependencies += "com.avast" %% "scala-server-toolkit-http4s-blaze-server" % "<VERSION>"`
+
+There are `http4s-*` modules that provide easy initialization of a server and a client. Http4s is an interface with multiple possible
+implementations - for now we provide only implementations based on [Blaze](https://github.com/http4s/blaze).
+
+Both server and client are configured via configuration `case class` which contains default values taken from the underlying implementations.
+Initialization of a server and a client is very simple if defaults are fine for you.
```
augi: This sentence IMHO doesn't bring any value so I would remove it.

example/src/main/scala/com/avast/server/toolkit/example/Test.scala
```diff
@@ -0,0 +1,43 @@
+package com.avast.server.toolkit.example
+
+object Test extends App {
```
augi: Do we need this file?

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClient.scala
sideeffffect: I don't object having a method which creates `SSLContext` behind the scene.
but at the same time, I think we should have a `make...` method which takes a custom `SSLContext` instance

jakubjanecek: Sure, that makes sense but let's add it once we need it. I don't want to prolong this PR anymore.

example/src/main/scala/com/avast/server/toolkit/example/Test.scala
jakubjanecek: No, that is a mistake, thanks.

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+final case class Http4sBlazeClientConfig(
+  responseHeaderTimeout: Duration = Duration.Inf,
+  idleTimeout: FiniteDuration = Duration(1, TimeUnit.MINUTES),
+  requestTimeout: FiniteDuration = defaults.RequestTimeout,
+  connectTimeout: FiniteDuration = defaults.ConnectTimeout,
+  userAgent: `User-Agent` = `User-Agent`(AgentProduct("http4s-blaze-client", Some(BuildInfo.version)), List(AgentComment("Server"))),
+  maxTotalConnections: Int = 10,
+  maxWaitQueueLimit: Int = 256,
+  maxConnectionsPerRequestkey: Int = 256,
+  sslContext: Option[SslContextConfig] = None,
```
sideeffffect: maybe it would be for the best, if `SslContextConfig` wasn't part of `Http4sBlazeClientConfig`, and was provided at the required places separately (either the `SslContextConfig` would be provided, or already a `SslContext` itself)

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/middleware/CorrelationIdMiddleware.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.slf4j.LoggerFactory
+
+import scala.language.higherKinds
+
+/** Provides correlation ID functionality. Either generates new correlation ID for a request or takes the one sent in HTTP header
+  * and puts it to [[org.http4s.Request]] attributes. It is also filled into HTTP response header.
+  *
+  * Use method `retrieveCorrelationId` to get the value from request attributes.
+  */
+class CorrelationIdMiddleware[F[_]: Sync](correlationIdHeaderName: CaseInsensitiveString,
```
sideeffffect: this class should implement some trait, so that in production code, we can program against that interface, and not this particular implementation
the good old, time proven part of dependency injection ;)

docs/http4s.md
jakubjanecek: Alright.

project/Dependencies.scala
```diff
@@ -2,16 +2,23 @@ import sbt._
 
 object Dependencies {
 
+  import Versions._
```
sideeffffect: this is not good in practice -- it's just a matter of time until the names will clash

http4s-blaze-client/src/main/scala/com/avast/server/toolkit/http4s/Http4sBlazeClientConfig.scala
jakubjanecek: As discussed personally let's keep it this way and extend if we ever need it.

http4s-blaze-server/src/main/scala/com/avast/server/toolkit/http4s/middleware/CorrelationIdMiddleware.scala
jakubjanecek: As discussed personally I do not see the need for that, middleware is a very specific thing.

project/Dependencies.scala
jakubjanecek: Right, I'll remove it.

PR #24, opened 2019-10-02
trailingCommas

sideeffffect:
life is easier, if trailing commas are everywhere
especially for git: history and rebasing
(it has happened to me sometimes that git rebase (or was it merge?) broke because of absence of commas)

the commas will be automatically added (IntelliJ will add them on every save), so this should be an effortless rule

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -9,3 +9,4 @@ align.openParenDefnSite = true
 rewrite.rules = [RedundantParens, SortImports, SortModifiers, PreferCurlyFors]
 includeNoParensInSelectChains = true
 lineEndings = preserve
+trailingCommas = always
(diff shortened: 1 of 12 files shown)


sideeffffect:
@augi @hanny24 @tomasherman 
I would like us to start using `trailingCommas`, I think it is a good practice
we can kicking off the discussion with this PR

jendakol:
I'm sorry but after I've gone through the code... TBH honest I'm confused by some pieces of code :smiley:  
I like the trailing comma for _varargs_ (multiline `Seq(... ...)`), I'm able to survive it in SBT config and imports, but adding trailing comma after last parameter of function - :no_entry_sign: please. It doesn't happen very often but now I agree with @jakubjanecek :wink:

hanny24:
My 2 cents: if we were to use scalastyle , we would end up with a bad time:( Maybe it's solved by newer version of scalastyle)

tomasherman:
i am not a fan of trailing commas - they make life a little bit easier but they add more noise and make code more confusing :( i don't like that scalafmt should be adding them everywhere automatically

i don't see how they make working with git easier - can you elaborate?

PR #30, opened 2019-10-07
feat: Add Monix and ZIO bundles, Micrometer and PureConfig interop

jakubjanecek:


jakubjanecek:
It doesn't necessarily must be like that but yes. This can be seen in the `pureconfig` module (package `implicits`) where `ConfigReader`s for other modules are implemented.

augi:
Maybe it would be better to start with README, so we could see your intents earlier?

micrometer/src/main/scala/com/avast/sst/micrometer/interop/Http4sServerMetrics.scala
```diff
@@ -0,0 +1,45 @@
...
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import com.avast.sst.micrometer.HttpStatusMetrics
+import io.micrometer.core.instrument.MeterRegistry
+import org.http4s.metrics.{MetricsOps, TerminationType}
+import org.http4s.{Method, Status}
+
+import scala.language.higherKinds
+
+class Http4sServerMetrics[F[_]: Sync](meterRegistry: MeterRegistry) extends MetricsOps[F] {
```
sideeffffect: this class contains implementation details which are incidental, and users shouldn't need to care about
the class can be made private and copanion object can have method `make` which will return just `MetricsOps`

pureconfig/src/main/scala/com/avast/sst/pureconfig/implicits/Http4sBlazeServer.scala
```diff
@@ -0,0 +1,14 @@
+package com.avast.sst.pureconfig.implicits
```
sideeffffect: we probably want to use the trick/technique from cats (& co.) where `implicits` is an object which has all implicit classes/convertors/... as members, so that you can then
```scala
import com.avast.sst.pureconfig.implicits._
```
and all implicit things are in scope

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  private val partNames = Vector("type", "scope", "name")
+
+  override def createName(`type`: String, domain: String, name: String): ObjectName = {
+    val parsedName = parseName(domain, name)
+    parsedName.getOrElse(defaultFactory.createName(`type`, domain, name))
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  private def parseName(domain: String, name: String): Try[ObjectName] = Try {
```
sideeffffect: `Try` (as everything Scala related from Twitter, like `Future` or hysterical community drama) come with more headache than worth it
It literally catches everything, which is almost always problematic. Could you perhaps catch only the expected exceptions (or `NonFatal`, or use `Either.catchNonFatal` or whatever it's called)?

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import com.codahale.metrics.jmx.{DefaultObjectNameFactory, ObjectNameFactory}
+import javax.management.ObjectName
+
+import scala.util.Try;
+
+/** This is custom [[com.codahale.metrics.jmx.ObjectNameFactory]] which uses "type-scope-name" hierarchy of resulting
+  * [[javax.management.ObjectName]] (levels 3-N are glued together).
+  */
+class TypeScopeNameObjectNameFactory(separator: String = ".") extends ObjectNameFactory {
```
sideeffffect: does this class need  to be public? can we just not return an instance of `ObjectNameFactory`?

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    parsedName.getOrElse(defaultFactory.createName(`type`, domain, name))
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  private def parseName(domain: String, name: String): Try[ObjectName] = Try {
+    val parts = name.split(quotedSeparator, partNames.length)
+
+    /* The following block of code is a little hack. The problem is that ObjectName requires HashTable as parameter but HashTable
+       is unsorted and thus unusable for us. We hack it by raping the HashTable and in-fact using LinkedHashMap which is
+       much more suitable for our needs. */
```
sideeffffect: why can't we just use `Hashtable`?

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
jakubjanecek: Not really, I can hide it from everyone. I just thought it could be useful to someone but you are right it's better to hide it.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
jakubjanecek: I don't think it could cause any real trouble here but yeah, I'll replace it with `Either`.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/TypeScopeNameObjectNameFactory.scala
jakubjanecek: It's described in the comment. By using `Hashtable` you would get unsorted list of entries but we need it sorted.

micrometer/src/main/scala/com/avast/sst/micrometer/interop/Http4sServerMetrics.scala
jakubjanecek: Right.

micrometer/src/main/scala/com/avast/sst/micrometer/interop/MicrometerHttp4sServerMetricsModule.scala
```diff
@@ -0,0 +1,21 @@
+package com.avast.sst.micrometer.interop
+
+import cats.effect.{Clock, Effect, Sync}
+import io.micrometer.core.instrument.MeterRegistry
+import org.http4s.HttpRoutes
+import org.http4s.server.middleware.Metrics
+
+import scala.language.higherKinds
+
+class MicrometerHttp4sServerMetricsModule[F[_]: Sync](val globalMetrics: HttpRoutes[F] => HttpRoutes[F],
```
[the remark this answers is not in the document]
jakubjanecek: You are right it could but modules are not supposed to be any kind of abstraction so I don't see the point of doing that.

pureconfig/src/main/scala/com/avast/sst/pureconfig/implicits/Http4sBlazeClient.scala
```diff
@@ -0,0 +1,29 @@
...
+  implicit val userAgentReader: ConfigReader[`User-Agent`] = ConfigReader[String].emap { value =>
+    `User-Agent`.parse(value).leftMap { parseFailure =>
+      CannotConvert(value, "User-Agent HTTP header", parseFailure.message)
+    }
+  }
+
+  implicit val parserModeReader: ConfigReader[ParserMode] = ConfigReader[String].map(_.toLowerCase).emap {
+    case "strict"  => Right(ParserMode.Strict)
+    case "lenient" => Right(ParserMode.Lenient)
+    case badValue  => Left(CannotConvert(badValue, "ParserMode", "strict|lenient"))
```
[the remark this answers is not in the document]
jakubjanecek: Thanks.

pureconfig/src/main/scala/com/avast/sst/pureconfig/implicits/Http4sBlazeServer.scala
jakubjanecek: Not in this case. Because the dependencies are optional you don't want to import everything at once. It would work if they were defined as `lazy val` but I'd rather force the user to make an explicit decision than to rely on this.

jakubjanecek:
@sideeffffect I've updated the code according to your comments.

@augi Yes, I guess I need to do that though I wanted to avoid that in case you would shred it to pieces :) If I could describe it in few sentences the point is that we have module per dependency (e.g. `http4s-blaze-server` and `pureconfig`) and these modules sometimes need to interoperate. My original idea was an idea of "bundle" which is a module that depends on both these dependencies and ties them together - we could have as many as we wanted of those with different combinations and with some duplication of code (though the duplication should be quite limited). Another option would be to create lots of small modules with different dependency combinations and only use them in "bundles" to share some code (to avoid bit of the duplication but with lots of extra work for us). And than you offered me the solution with optional dependencies which I don't like because of possible runtime dependencies (and I cannot really always use the trick with failing user's code by forcing him to provide one of the classes from the dependency) but in the end I think it offers me the best of both worlds which is that I keep implementation of interop code close the the main dependency (e.g. implementation of `ConfigReader` from PureConfig for all other modules inside the `pureconfig` module but with only *optional* dependency on other modules) but I force the user to make an explicit import (with proper import name) to be aware that something special is used (if you're importing `com.avast.sst.pureconfig.implicits.Http4sBlazeClient._` you should probably realize that you need to have `sst-http4s-blaze-client` dependency in your project). I also presume that "bundles" which will also be used to bundle certain combinations of dependencies should prevent most of runtime exceptions because they will already bundle together necessary dependencies.

jakubjanecek:
I'll work on this tomorrow to add some documentation and examples. I'll make it a proper full-blown PR which you can evaluate.

hanny24:
So my take on this is to avoid optional dependencies and make many small modules instead. First of all, I believe that creating a new module is fairly easy and does not really take much time.
Secondly, I think that many small modules is a good practice as we can see in many scala libraries: for instance take a look [here](https://pureconfig.github.io/docs/library-integrations.html) or [here](https://circe.github.io/circe/api/).

And lastly, this solution is quite scaleable as you should only integrate only *two* libraries together.

jakubjanecek:
I kind of agree because that was my original intent and the most "pure" solution. However the situation is a bit different with this library. What you refer to is a single library interoperating with N external libraries which results in N SBT modules. What we have is M SBT submodules interoperating with N external libraries which results in theory in M times N SBT submodules. Right now we have PureConfig interoperating with 5 other libraries so we would have to create 5 other submodules. For example we'd have to create two submodules for Micrometer - one for JMX and one for StatsD implementation. Now imagine that in a year we'll replace or add (more probable) new configuration library "FutureConfig"... you have to double the number of submodules immediately. And it gets worse, we now support single implementation of http4s but I can very easily imagine adding support for different implementations, e.g. supporting HTTP client implemented using Apache. So yeah, creating a submodule is easy but having tens of modules with usually only single file seems quite extreme. Thoughts?

hanny24:
I'm afraid I don't see how Micrometer is related configuration library. Say I have http4s. I would expect following modules:
- `sst-http4s`
- `sst-http4s-blaze`
- `sst-http4s-apache`
- `sst-http4s-micrometer`
- `sst-http4s-pureconfig`

For me this is very straight forward, clean and manageable. My point is that optional dependencies is just an excuse for our laziness that will hurt us in the future. BTW we already tried something like this before with `typeclasses` library that we have and it did not worked and we had to split it.

jakubjanecek:
Can you maybe try to explain more what was the problem you encountered?

I get what you propose and again I kind of agree. But we do have `http4s-blaze-server` and `http4s-blaze-client` so you'd have to double these and the way Micrometer and PureConfig are "related" is that we need to configure Micrometer and we do that by some case class and we need to load that case class from Lightbend Config. Look at the code, it's all there: https://github.com/avast/scala-server-toolkit/pull/30/files#diff-c23502bc80a3b79260a553e543fd78a9

Also I think that you have the naming in reverse. `sst-http4s-pureconfig` somehow evokes that it's http4s' implementation for PureConfig whereas what we have is PureConfig's implementation for `http4s` and thus it should be `sst-pureconfig-http4s-blaze` at least (the same it works here: https://pureconfig.github.io/docs/library-integrations.html).

So this is what we would need now (if I am not forgetting something):
* `sst-http4s-blaze-server`
* `sst-http4s-blaze-client`
* `sst-pureconfig`
* `sst-pureconfig-http4s-blaze-server`
* `sst-pureconfig-http4s-blaze-client`
* `sst-micrometer`
* `sst-micrometer-jmx`
* `sst-pureconfig-micrometer-jmx`
* `sst-micrometer-http4s-blaze-server`
* `sst-micrometer-http4s-blaze-client`

And just by adding new Micrometer implementation (StatsD) and possibly new http4s implementation (Apache) you can immediately add like another 8 or maybe 10 submodules. To me it seems like it's not scalable at all and will get out of hand quite quickly.

The problem is that our library does not provide anything new, it's just a mix of other libraries so dependencies might be the biggest problem we're dealing with. That's why I also think that copying others is not the best argument.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
```diff
@@ -0,0 +1,52 @@
...
+    JmxReporter
+      .forRegistry(metricRegistry)
+      .inDomain(domain)
+      .createsObjectNamesWith(new TypeScopeNameObjectNameFactory())
+      .build
+  }
+
+  private class DomainJmxConfig(domain: String) extends JmxConfig {
+    @SuppressWarnings(Array("org.wartremover.warts.Null"))
+    override def get(key: String): String = null
```
jendakol: I strongly vote for adding a comment why `null` is _legal_ here. It always kinda smells...

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
```diff
@@ -0,0 +1,52 @@
...
+    override def domain(): String = domain
```
jendakol: `val domain: String`? :thinking:

jakubjanecek:
Jenda points out an important fact. If I make all the PureConfig reader definitions `implicit lazy val` it should never happen that `NoClassDefFound` is thrown because unless you have the type on the classpath you will never try to get the `implicit` definition for it, you would have to explicitly reference it which would be very strange.

Micrometer module is a bit different but I think I do what Augi told me on Slack which is that I require some class from the dependency which again prevents you from the runtime exception because your code wouldn't compile.

Please consider that because I think it makes the optional dependencies more digestible ;)

hanny24:
> Can you maybe try to explain more what was the problem you encountered?

- Runtime errors as it was not very clear when things get accessed/initialized. 
- It wasn't very user friendly as it was difficult to track where is what implemented (eg. if configuration via pureconfig is implemented as optional dependency on pureconfig in `http4s` module, or optional dependency on `http4s` in `pureconfig` module)

> I get what you propose and again I kind of agree. But we do have `http4s-blaze-server` and `http4s-blaze-client` so you'd have to double these and the way Micrometer and PureConfig are "related" is that we need to configure Micrometer and we do that by some case class and we need to load that case class from Lightbend Config. Look at the code, it's all there: https://github.com/avast/scala-server-toolkit/pull/30/files#diff-c23502bc80a3b79260a553e543fd78a9

Could you please be more specific what I'm supposed to look at? I still don't see why would you need two modules for http4s and micrometer integration. Just define metrics trait in a module (or reuse `MetricsOps` in case of http4s) and provide micrometer implementation in a separate module.

> Also I think that you have the naming in reverse. `sst-http4s-pureconfig` somehow evokes that it's http4s' implementation for PureConfig whereas what we have is PureConfig's implementation for `http4s` and thus it should be `sst-pureconfig-http4s-blaze` at least (the same it works here: https://pureconfig.github.io/docs/library-integrations.html).

Well, that's subjective:) But I don't really care about this as long as it's consistent.

> 
> So this is what we would need now (if I am not forgetting something):
> 
>     * `sst-http4s-blaze-server`
>     * `sst-http4s-blaze-client`
>     * `sst-pureconfig`
>     * `sst-pureconfig-http4s-blaze-server`
>     * `sst-pureconfig-http4s-blaze-client`
>     * `sst-micrometer`
>     * `sst-micrometer-jmx`
>     * `sst-pureconfig-micrometer-jmx`
>     * `sst-micrometer-http4s-blaze-server`
>     * `sst-micrometer-http4s-blaze-client`
> 
> 
> And just by adding new Micrometer implementation (StatsD) and possibly new http4s implementation (Apache) you can immediately add like another 8 or maybe 10 submodules. To me it seems like it's not scalable at all and will get out of hand quite quickly.
> 

But then we won't really fulfill one of our goals: to have small building blocks. We will end up with a few big ones. And if I take into extreme, we can have single module only.

jveselka:
> Also I think that you have the naming in reverse. sst-http4s-pureconfig somehow evokes that it's http4s' implementation for PureConfig whereas what we have is PureConfig's implementation for http4s and thus it should be sst-pureconfig-http4s-blaze at least (the same it works here: https://pureconfig.github.io/docs/library-integrations.html).

I don't think that would be very practical. You usually ask yourself "How can I improve my `sst-http4s-blaze-server` with additional modules?" (by looking for modules with prefix `sst-http4s-blaze-server`). Asking yourself "What more can I configure with `sst-pureconfig`" would be strange.

jakubjanecek:
@hanny24 I have to think about your message so I'll respond later.

@jveselka Get it but that leads directly to the "excessive number of submodules" state so in a way the reverse naming (though probably less natural) could work better.

Let me ask you this. *Is there anyone who thinks optional dependencies are a viable/better/simpler/whatever solution?*

(it seems like Jenda thinks so but if everyone else is against it I guess there's no point in discussing this more)

We could also try to help us and make less granular submodules. For example I wonder whether `http4s-server` and `http4s-client` need to be split because usually you want both. I also split the `jvm-*` modules into multiple ones (asked by @jveselka) but since they have the same dependency which is JVM I wonder whether we could put them back together to help ourselves a bit. I am afraid of getting unnecessary dependencies into my project, not unnecessary small pieces of code (such as SSL context configuration which I usually don't use at all). What do you say?

hanny24:
I'm ok with merge modules as long as all of them require same set of dependencies. It the merged module turns out to be big, we can always split it later.

tomasherman:
So for me if the goal is to have _unopinionated_ library for composing small pieces together into a backend, multiple small modules is the way to. Now we should discuss whether _unopinionated_ is really a way to go. 

The tradeoffs are:
#### Optional dependencies
 - 👍 less work upfront when creating `sst` and user application
 - 👎 possibility of runtime errors (and even nasty ones in cases when some class is not needed until hours after deployment and it gets tricky to even detect that something bad is going on)
#### Modules
- 👍 more work upfront
- 👎 mode headache at runtime

Neither of those choices is very appealing. Which makes me question whether having unopinionated toolbox is a way to go.

Btw https://git.int.avast.com/skyline/backend-common can be used as example of what library with many small modules looks like - it's up to you to decide whether it's worth it to have it structured this way (and our approach is much more opinionated that this library is going to be).

sideeffffect:
my 2 haléře:

 * if we wanted to be failing in run-time, we could be already using Python, so if the work is not absurdly demanding, I would be for having more smaller modules
 * agree with @jveselka , I like the "added fature" as the last thing in the name, so `sst-http4s-blaze-client-pureconfig`, or `sst-http4s-blaze-server-micrometer`.
 * maybe the names should be even `sst-http4s-client-blaze-pureconfig`, or `sst-http4s-server-blaze-micrometer`, because there are other http4s servers and clients besides the ones based on `blaze`

jakubjanecek:
Thank you everyone for your comments. I've decided to get rid of optional dependencies and reworked it into multiple smaller modules (on the other hand I merged all the `jvm-*` modules because it was unnecessary to have them split). I still need to work on some documentation and examples and maybe tests but please have a look at it whether it now makes sense to you.

hanny24:
Could you please split it into multiple smaller PRs? It's way to big.

jakubjanecek:
Sorry but I cannot spend more time on redoing this. I know there's a lot of changes and I will do better next time but as it's all inter-related it would be quite difficult to commit the changes one-by-one and not spend another day or two on this. Please bear with me and try to have a look at it.

The most important is whether the artifact names, package names and the way implicits are imported (PureConfig) makes sense to you. Other changes are usually just cosmetic or moving of files from the past. I also added Micrometer monitoring but you've also seen that in the past and last is two bundles for Monix and ZIO which just add `Server` trait (just a helper) and bundle dependencies together.

jakubjanecek:
I've added some tests, comments, documentation and most importantly fixed the build (caused by case-sensitivity of the file system on Travis). This PR is ready to be merged from my PoV.

http4s-client-blaze-pureconfig/src/main/scala/com/avast/sst/http4s/client/pureconfig/package.scala
```diff
@@ -0,0 +1,3 @@
+package com.avast.sst.http4s.client
+
+package object pureconfig extends ConfigReaders
```
sideeffffect: shouldn't the name of the object which contains implicits be `implicits`, not `pureconfig` (as is the custom in the TypeLevel world)?

http4s-server-micrometer/src/main/scala/com/avast/sst/http4s/server/micrometer/MicrometerHttp4sMetricsOpsModule.scala
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.TimeUnit
+
+import cats.effect.Sync
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import io.micrometer.core.instrument.MeterRegistry
+import org.http4s.metrics.{MetricsOps, TerminationType}
+import org.http4s.{Method, Status}
+
+object MicrometerHttp4sMetricsOpsModule {
```
sideeffffect: I'm wondering whether we need this `*Module` suffix everywhere :thinking: 
but thankfully this is not the most burning question

http4s-client-blaze-pureconfig/src/main/scala/com/avast/sst/http4s/client/pureconfig/package.scala
jakubjanecek: You mean `pureconfig.implicits._` or just `implicits._`. If it was just `implicits._` I wonder what happens if I have two modules with the same `package object`? Does that even work?

sideeffffect: in this case, I think it should be `com.avast.sst.http4s.client.pureconfig.implicits`. Then the use would do
```scala
import com.avast.sst.http4s.client.pureconfig.implicits._
```
does that make sense?

http4s-server-micrometer/src/main/scala/com/avast/sst/http4s/server/micrometer/MicrometerHttp4sMetricsOpsModule.scala
sideeffffect: image

sideeffffect:
so where do you @jakubjanecek want that green thumb exactly?
image

bundle-monix-http4s-blaze/src/main/scala/com/avast/sst/bundle/MonixServerApp.scala
```diff
@@ -0,0 +1,36 @@
...
+import monix.eval.{Task, TaskApp}
+import org.http4s.server.Server
+import org.slf4j.LoggerFactory
+
+/** Extend this `trait` if you want to implement server application using [[monix.eval.Task]] effect data type.
+  *
+  * Implement method `program` with initialization and business logic of your application. It will be automatically run until JVM is shut
+  * down in which case all the resources are cleaned up because the whole `program` is a [[cats.effect.Resource]].
+  */
+trait MonixServerApp extends TaskApp {
```
augi: It looks like this project is kind of example, right? Would be better to have it just shown in a README? It's scary how many modules we have here...

jakubjanecek: No, this is not an example, this is a helper `trait` to simplify server initialization. And the submodule in itself "bundles" together certain dependencies which is the more important feature.

augi: Yes, but this requires that the program must have at least one HTTP server based on Blaze.

It would make more sense for me having a module that would contain a `MonixResourceApp` base class/trait that would have `A` instead of `org.http4s.server.Server`.

And the same for `ZioServerApp` - I would prefer to have _generic_ `ZioResourceApp` instead.

jakubjanecek: Good point, I guess I'll add it but the problem will be that the server app automatically "blocks" whereas you don't want to do that for generic `A`. So I think we need both. I've done this one automatically because the project is called "server" so it seemed natural to provide that.

PR #31, opened 2019-10-10
feat: Add *ResourceApp trait to allow more generic applications

jakubjanecek:
I am not sure about the use of this, I think it's quite limited. But it does not hurt to have it so if you think it's good to have it we can merge it.


augi:
Please no, keep it as is.

PR #44, opened 2019-10-18
feat: Add ssl-config module

jakubjanecek:
BREAKING CHANGE: SslContextModule from jvm was replaced by the new one


ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
```diff
@@ -0,0 +1,26 @@
+package com.avast.sst.ssl
```
sideeffffect: the directory should be named `ssl-context` (not `ssl-config`)

jakubjanecek: No, it's based on the dependency which is named `ssl-config`. If anything I could rename the package to `sslconfig` or something like that but I though just `ssl` would work better.

sideeffffect: :/ sorry, I'm probably confused... the point of this (sub)project is to create an instance of `javax.net.ssl.SSLContext`, no? that's why I would expect the name to be `ssl-context`

jakubjanecek: Yes but the names of the modules are based on the dependencies, not the contens of the module (though these are very related usually ;) ). Theoretically there could be multiple implementations of how to create `SSLContext` so if I name it just `sst-ssl-context` I wouldn't know which dependency/implementation it is and if I wanted to have both in this project I would have a conflict. I understand your point but this way it is extensible and consistent with the rest of the modules.

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
```diff
@@ -0,0 +1,26 @@
...
+  SSLConfigFactory
+}
+import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
+
+import scala.language.higherKinds
+
+object SslContextModule {
+
+  /** Initializes [[javax.net.ssl.SSLContext]] from the provided config. */
+  def make[F[_]: Sync](config: Config): F[SSLContext] = Sync[F].delay {
```
sideeffffect: now, this might come across as way too pedantic, but: I like that we don't use `com.typesafe.config.Config` in our "API". And I would prefer if it stayed that way.

The fact that we're creating the `SSLContext` using `com.typesafe.sslconfig.ssl` is an implementation detail, which be leaking out.

suggestion: create a config case class (as everywhere else) to configure SSLContext and change `make` method to use that

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
sideeffffect: you're absolutely on point that this is just one way to create a `SSLContext`, and there could be many other, at least in theory. Maybe the subproject name should reflect that: `sslcontext-typesafe`?

jakubjanecek: The subproject reflects that because the library is called `ssl-config`: https://github.com/lightbend/ssl-config

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
jakubjanecek: Personally we've discussed this and I decided to keep the name because it reflects the dependency used.

PR #46, opened 2019-10-21
Add doobie-hikari module

jakubjanecek:


doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,44 @@
...
+object DoobieHikariModule {
+
+  /** Makes [[doobie.hikari.HikariTransactor]] initialized with the given config.
+    *
+    * @param boundedConnectExecutionContext [[scala.concurrent.ExecutionContext]] used for creating connections (should be bounded!)
+    */
+  def make[F[_]: Async](config: DoobieHikariConfig,
+                        boundedConnectExecutionContext: ExecutionContext,
+                        blocker: Blocker,
+                        metricsTrackerFactory: MetricsTrackerFactory)(implicit cs: ContextShift[F]): Resource[F, HikariTransactor[F]] = {
```
sideeffffect: to make things general, let's make `MetricsTrackerFactory` optional, since it doesn't have to be set

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,44 @@
...
+                         c.setUsername(config.username)
+                         c.setPassword(config.password)
+                         c.setAutoCommit(config.autoCommit)
+                         c.setConnectionTimeout(config.connectionTimeout.toMillis)
+                         c.setIdleTimeout(config.idleTimeout.toMillis)
+                         c.setMaxLifetime(config.maxLifeTime.toMillis)
+                         c.setMinimumIdle(config.minimumIdle)
+                         c.setMaximumPoolSize(config.maximumPoolSize)
+                         c.setReadOnly(config.readOnly)
+                         c.setLeakDetectionThreshold(config.leakDetectionThreshold.map(_.toMillis).getOrElse(0))
```
sideeffffect: shouldn't this be more like
```scala
config.leakDetectionThreshold.map(t => c.setLeakDetectionThreshold(t.toMillis))
```
:thinking:

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariConfig.scala
```diff
@@ -0,0 +1,18 @@
...
+                                    username: String,
+                                    password: String,
+                                    autoCommit: Boolean = true,
+                                    connectionTimeout: FiniteDuration = FiniteDuration(30, TimeUnit.SECONDS),
+                                    idleTimeout: FiniteDuration = FiniteDuration(10, TimeUnit.MINUTES),
+                                    maxLifeTime: FiniteDuration = FiniteDuration(30, TimeUnit.MINUTES),
+                                    minimumIdle: Int = 10,
+                                    maximumPoolSize: Int = 10,
+                                    readOnly: Boolean = false,
+                                    leakDetectionThreshold: Option[FiniteDuration] = None)
```
sideeffffect: or shouldn't this be
```scala
leakDetectionThreshold: FiniteDuration = FiniteDuration.Zero // or however you describe 0
```
:thinking: :thinking: :thinking:

example/src/main/scala/com/avast/sst/example/service/RandomService.scala
```diff
@@ -0,0 +1,25 @@
...
+
+trait RandomService {
+
+  def randomNumber: Task[Double]
+
+}
+
+object RandomService {
+
+  def apply(transactor: Transactor.Aux[Task, _]): RandomService = new RandomService {
```
sideeffffect: classes are for loosers! :heart: :muscle:

sideeffffect: why not just `HikariTransactor[Task]`?

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariConfig.scala
jakubjanecek: It could be but I like `Option` more, that's why I am converting it later.

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
jakubjanecek: Right.

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
jakubjanecek: Right.

example/src/main/scala/com/avast/sst/example/service/RandomService.scala
jakubjanecek: I don't think the service needs to know about the implementation of `Transactor`.

sideeffffect: Oh that's right! good point.
in that case would `Transactor[Task]` also work? (`Transactor.Aux[Task, _]` is a bit ugly...)

docs/micrometer.md
```diff
@@ -48,10 +48,10 @@ for {
 // res0: Resource[Task, Server[Task]] = Bind(
 //   Bind(
 //     Bind(
-//       Suspend(zio.ZIO$FlatMap@34ede267),
-//       com.avast.sst.jvm.execution.ExecutorModule$$$Lambda$1652/815927219@6b92a0d1
+//       Suspend(zio.ZIO$FlatMap@56554365),
+//       com.avast.sst.jvm.execution.ExecutorModule$$$Lambda$1652/1769513770@2f2dc407
 //     ),
-//     cats.effect.Resource$$Lambda$1653/888293905@51e0f2eb
+//     cats.effect.Resource$$Lambda$1653/1432768646@320770d7
```
sideeffffect: is there a way to prevent these useless changes in the future?

tomasherman: https://scalameta.org/mdoc/docs/modifiers.html#silent ?

tomasherman: pls use it - this is horrible

docs/doobie.md
```diff
@@ -0,0 +1,40 @@
...
+  executorModule <- ExecutorModule.makeFromExecutionContext[Task](runtime.Platform.executor.asEC)
+  meterRegistry <- MicrometerJmxModule.make[Task](configuration.jmx)
+  boundedConnectExecutionContext <- executorModule
+                                     .makeThreadPoolExecutor(
+                                       configuration.boundedConnectExecutor,
+                                       new ConfigurableThreadFactory(Config(Some("hikari-connect-%02d")))
+                                     )
+                                     .map(ExecutionContext.fromExecutorService)
+  hikariMetricsFactory = new MicrometerMetricsTrackerFactory(meterRegistry)
+  doobieTransactor <- DoobieHikariModule
```
tomasherman: can you pls provide some documentation why hikari was chosen? I never heard about this

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,46 @@
...
+  def make[F[_]: Async](
+    config: DoobieHikariConfig,
+    boundedConnectExecutionContext: ExecutionContext,
+    blocker: Blocker,
+    metricsTrackerFactory: Option[MetricsTrackerFactory] = None
+  )(implicit cs: ContextShift[F]): Resource[F, HikariTransactor[F]] = {
+    for {
+      hikariConfig <- Resource.liftF {
+                       Sync[F].delay {
+                         val c = new HikariConfig()
```
tomasherman: i would move this into it's own function that can be pure and not wrapped in `F`

docs/doobie.md
jakubjanecek: I think it's the industry standard for SQL connection pool, not much to "document". Am I right @sideeffffect?

tomasherman: i doubt it's `industry standart` :D but im not against it - just would like to see at least some rationalization since i never heard about this library nor have ever seen it used

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
jakubjanecek: I moved it to its own function but since this is Java code I cannot believe it's not doing some side effects so I definitely need to wrap it in F. You can actually see it might try to load some properties from file so just to be sure I will wrap it.

docs/doobie.md
sideeffffect: It's the only connection pool that's documented on Dobbie's website, so I guess vast majority of Doobie users use Hikari :laughing: 
https://tpolecat.github.io/doobie/docs/14-Managing-Connections.html#using-a-hikaricp-connection-pool

jakubjanecek: How can you doubt it when you don't know anything about it? I think you get mislead by the name which seems strange but it really is a very popular JDBC connection pooling library. Sure, frameworks like Spring or J2EE have probably their own solution but in the J2SE-land Hikari is definitely in the top 3. Second link from Google search mentions it: https://www.baeldung.com/java-connection-pooling Another argument which @sideeffffect told me is that doobie does not support anything else so there's not much choice ;) But I guess author of doobie chose it for some reason, right?

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,45 @@
...
+  def make[F[_]: Async](
+    config: DoobieHikariConfig,
+    boundedConnectExecutionContext: ExecutionContext,
+    blocker: Blocker,
+    metricsTrackerFactory: Option[MetricsTrackerFactory] = None
+  )(implicit cs: ContextShift[F]): Resource[F, HikariTransactor[F]] = {
+    for {
+      hikariConfig <- Resource.liftF(Sync[F].delay(makeHikariConfig(config, metricsTrackerFactory)))
+      transactor <- HikariTransactor.fromHikariConfig(hikariConfig, boundedConnectExecutionContext, blocker)
+    } yield transactor
```
sideeffffect:
```scala
    val hikariConfig = makeHikariConfig(config, metricsTrackerFactory)
    for {
      transactor <- HikariTransactor.fromHikariConfig(hikariConfig, boundedConnectExecutionContext, blocker)
    } yield transactor 
```

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  private def makeHikariConfig(config: DoobieHikariConfig, metricsTrackerFactory: Option[MetricsTrackerFactory]) = {
```
sideeffffect: maybe to be super purist, since we want to assume this could have sideeffects, maybe the return type should be `F[HikariConfig]` :thinking: 

I know the method is private -- but it already confused me a bit -- it's good to use types even for documentation and explaining intent

sideeffffect: https://github.com/openbouquet/HikariCP/blob/master/src/main/java/com/zaxxer/hikari/HikariConfig.java#L113

it truly is side-effectful

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariConfig.scala
```diff
@@ -0,0 +1,18 @@
+package com.avast.sst.doobie
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.FiniteDuration
+
+final case class DoobieHikariConfig(driver: String,
```
sideeffffect: and to be super duper faithful to the original config, shouldn't we add the missing fields?

https://github.com/openbouquet/HikariCP/blob/master/src/main/java/com/zaxxer/hikari/HikariConfig.java#L58

docs/doobie.md
tomasherman: ok, fair enough ... it seemd like we are using some opt-in functionality ... do we need to call it `doobiehikari` module then? could we call it just `DoobieModule`?

jakubjanecek: Well, I look into the future and expect that there could be more implementations. This is consistent with the way `http4s` modules are structured etc.

tomasherman: i suspected that is the case - i generally dislike polliuting api with vision that there can be some different implementation eventually (even though it's been years and non came up and the default is *industry standart*) i think it's a case of premature optimization ... but we probably not gonna find an agreement on this issue so do what you wish

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariConfig.scala
jakubjanecek: Done.

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+                                    maximumPoolSize: Int = 10,
+                                    readOnly: Boolean = false,
+                                    leakDetectionThreshold: Option[FiniteDuration] = None,
+                                    allowPoolSuspension: Boolean = false,
+                                    initializationFailTimeout: Option[FiniteDuration] = None,
+                                    isolateInternalQueries: Boolean = false,
+                                    poolName: Option[String] = None,
+                                    registerMBeans: Boolean = false,
+                                    validationTimeout: Option[FiniteDuration] = None,
+                                    transactionIsolation: Option[TransactionIsolation])
```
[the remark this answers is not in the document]
jakubjanecek: 👍

jakubjanecek:
@sideeffffect @tomasherman Thank you for your review. I was in a hurry for some reason and made a lot of mistakes so thank you for catching them.

doobie-hikari/src/main/scala/com/avast/sst/doobie/DoobieHikariModule.scala
```diff
@@ -0,0 +1,76 @@
...
+  /** Makes [[doobie.hikari.HikariTransactor]] initialized with the given config.
+    *
+    * @param boundedConnectExecutionContext [[scala.concurrent.ExecutionContext]] used for creating connections (should be bounded!)
+    */
+  def make[F[_]: Async](
+    config: DoobieHikariConfig,
+    boundedConnectExecutionContext: ExecutionContext,
+    blocker: Blocker,
+    metricsTrackerFactory: Option[MetricsTrackerFactory] = None
+  )(implicit cs: ContextShift[F]): Resource[F, HikariTransactor[F]] = {
```
sideeffffect: maybe we should return `Transactor.Aux[M, DataSource]` (instead of `HikariTransactor[F]` == `Transactor.Aux[M, HikariDataSource]`)
:thinking:

jakubjanecek: I don't think so. You can always "upcast" it to that but you can never "downcast" it. I think the "rule" is to accept the most generic thing and return the most specific thing. Returning `HikariTransactor` allows the user to use something specific about it (just in case they need it).

sideeffffect: yeah, that's totally my question here actually: do we want the downstream users to "know" this is not a Transactor on just any `DataSource`, but it is on `HikariDataSource` specifically?

maybe that's an implementation detail we want to abstract away :smirk:

reading from the sources, the Hikari specific members are: `HikariPoolMXBean getHikariPoolMXBean()`, `HikariConfigMXBean getHikariConfigMXBean()` and `void evictConnection(Connection connection)`

jakubjanecek: I do think we don't want to abstract it. Initialization modules are not an abstraction, they just help you initialize and put stuff together. I think you want to get the specific type because what you will do you will use it as `Transactor[F]` anyway but just in case you needed to call one of those methods you have the option in your `main` method. I think this is really good and unharmful strategy that can be applied without thinking.

PR #53, opened 2019-10-21
Add Flyway support

jakubjanecek:


flyway/src/main/scala/com/avast/sst/flyway/FlywayModule.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.flywaydb.core.Flyway
+
+object FlywayModule {
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def make[F[_]: Sync](config: FlywayConfig): F[Flyway] = {
+    Sync[F].delay {
+      val builder = Flyway
+        .configure
+        .dataSource(config.url, config.username, config.password)
```
sideeffffect: shouldn't we prefer `dataSource​(javax.sql.DataSource dataSource)` (to `dataSource​(java.lang.String url, java.lang.String user, java.lang.String password)`)?

we could then use `Transactor.kernel` as the `javax.sql.DataSource` instance

I think it's generally a good practice to use objects instead of string configuration like usernames, passwords, paths to files as strings and so on

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
```diff
@@ -0,0 +1,24 @@
+package com.avast.sst.flyway
+
+final case class FlywayConfig(url: String,
+                              username: String,
+                              password: String,
+                              baselineOnMigrate: Boolean = false,
+                              baselineVersion: Option[String] = None,
```
sideeffffect: could we use `org.flywaydb.core.api.MigrationVersion` instead of String?

flyway/src/main/scala/com/avast/sst/flyway/FlywayModule.scala
```diff
@@ -0,0 +1,40 @@
...
+import cats.effect.Sync
+import org.flywaydb.core.Flyway
+
+object FlywayModule {
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def make[F[_]: Sync](config: FlywayConfig): F[Flyway] = {
+    Sync[F].delay {
+      val builder = Flyway
+        .configure
```
sideeffffect: just for info: in case the fluent config turns out cumbersome, there's still a [classical one](https://flywaydb.org/documentation/api/javadoc/org/flywaydb/core/api/configuration/ClassicConfiguration.html)

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
```diff
@@ -0,0 +1,24 @@
...
+                              baselineDescription: Option[String] = None,
+                              batch: Boolean = false,
+                              cleanDisabled: Boolean = false,
+                              cleanOnValidationError: Boolean = false,
+                              connectRetries: Int = 0,
+                              encoding: String = "UTF-8",
```
sideeffffect: again, shouldn't we prefer `java.nio.charset.Charset` here? :thinking:

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
jakubjanecek: I agree with `Charset` but `MigrationVersion` is really just a wrapper around `String`, it nevers fails , it wraps any string, so I don't see the benefit of having to create a `ConfigReader` for this because no validation will performed, what do you think?

flyway/src/main/scala/com/avast/sst/flyway/FlywayModule.scala
jakubjanecek: OK, let's do `DataSource`.

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
sideeffffect: :thinking: 
hmm, you're right that it's not super strongly typed, but still, you can use [`MigrationVersion.LATEST`](https://github.com/flyway/flyway/blob/6e2e4ed16651bcb0c15366863a1912a8c0574942/flyway-core/src/main/java/org/flywaydb/core/api/MigrationVersion.java#L37)

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
```diff
@@ -0,0 +1,25 @@
+package com.avast.sst.flyway
+
+import java.nio.charset.{Charset, StandardCharsets}
+
+import org.flywaydb.core.api.MigrationVersion
+
+final case class FlywayConfig(baselineOnMigrate: Boolean = false,
+                              baselineVersion: Option[MigrationVersion] = None,
```
sideeffffect: shouldn't the default be `EMPTY`? (I have no idea, don't really understand the issue at hand, just asking questions :) )

jakubjanecek: Actually the default is "1" but I don't want to have that in the config so I decided to make it optional and not set it in this case.

sideeffffect: sice we have `baselineVersion: Option[MigrationVersion] = None`, maybe we should also have something like
```scala
target: Option[MigrationVersion] = None
...
.setTarget(target)
```

flyway-pureconfig/src/main/scala/com/avast/sst/flyway/pureconfig/ConfigReaders.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.avast.sst.flyway.FlywayConfig
+import org.flywaydb.core.api.MigrationVersion
+import pureconfig.ConfigReader
+import pureconfig.error.ExceptionThrown
+import pureconfig.generic.semiauto.deriveReader
+
+trait ConfigReaders {
+
+  implicit private[pureconfig] val charsetReader: ConfigReader[Charset] = ConfigReader[String].emap { value =>
+    try {
```
sideeffffect: why not `Either.catchNonFatal`?

jakubjanecek: ok

flyway/src/main/scala/com/avast/sst/flyway/FlywayConfig.scala
jakubjanecek: ok

PR #56, opened 2019-10-23
feat: Add monix-catnap module with PureConfig integration

### LINKED ISSUE #50: Configuration case class for Monix CircuitBreaker
Monix provides `CircuitBreaker` but we should be able to configure it using case class which can be loaded from configuration file.

Some interop module between Monix and PureConfig would be good.

jakubjanecek:
Closes #50


monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerConfig.scala
```diff
@@ -0,0 +1,8 @@
+package com.avast.sst.monix.catnap
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+final case class CircuitBreakerConfig(maxFailures: Int,
+                                      resetTimeout: FiniteDuration,
+                                      exponentialBackoffFactor: Double = 1.0,
```
sideeffffect: I have 0 experience with Monix's circuit breaker, but just from the name, `exponentialBackoffFactor`, isn't value `1.0` completely non-sensical?

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,14 @@
+package com.avast.sst.monix.catnap
+
+import cats.effect.{Clock, Sync}
+import monix.catnap.CircuitBreaker
+
+object CircuitBreakerModule {
+
+  /** Makes [[monix.catnap.CircuitBreaker]] initialized with the given config and optionally [[cats.effect.Clock]]. */
+  def make[F[_]: Sync](config: CircuitBreakerConfig, clock: Option[Clock[F]] = None): F[CircuitBreaker[F]] = {
+    implicit val finalClock: Clock[F] = clock.getOrElse(Clock.create[F])
```
sideeffffect: I don't know if it's a good practice to create a `Clock` instance here. Don't you think we should just "force" the user to provide it always?

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerConfig.scala
jakubjanecek: This is the default. I guess the author of Monix decided that 1 is a good starting value.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: Why? `Clock` gives you two methods, `monotonic` and `realTime`, so I actually think we could even simplify it and not let users pass in `Clock` but just for testing  I allow it.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerConfig.scala
sideeffffect: fair enough

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
sideeffffect: inversion of control? things with functionality, even `Clock` shouldn't be created inside functions which need them, but should be passed from above.

sooner or later you will regret if you don't do it that way :) the only thing I'm not sure of is how to pass it, there are more options
```scala
def make[F[_]: Sync: Clock](config: CircuitBreakerConfig])
```
```scala
def make[F[_]: Sync](config: CircuitBreakerConfig)(implicit clock: Clock[F])
```
```scala
def make[F[_]: Sync](config: CircuitBreakerConfig, clock: Clock[F])
```

tomasherman: `def make[F[_]: Sync: Clock](config: CircuitBreakerConfig])`

jakubjanecek: But that's what I am saying.... the only place you would need that is testing the circuit breaking which I doubt anyone will do. So my point is, I allow you to pass in your clock if you need to but most people will not care so I create a new one (which is identical to any you could create yourself because this type of clock is not timezone or anything like that specific). So we kind of get best of both worlds by using `Option`. I wouldn't pass it implicitly, so the only thing that could be changed (I don't agree because it will be much worse for the users) is to make it non-optional.

tomasherman: def don't think it's a good idea to create it in here

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,14 @@
...
+    CircuitBreaker.of[F](config.maxFailures, config.resetTimeout, config.exponentialBackoffFactor, config.maxResetTimeout)
```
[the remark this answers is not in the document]
jakubjanecek: Tomas, you're mixing internal stuff with public OSS project :) There is no public circuit breaker implementation with `F` that I am aware of so this is our only option right now. I know this "module" does not provide much but the whole point is to provide "the same feel" to everything which means we need to have a configuratio case class and provide PureConfig readers so that it can be used together with the rest of the toolkit.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: Could you maybe be more specific? Have you read my extensive explanations? :)

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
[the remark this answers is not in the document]
tomasherman: > Tomas, you're mixing internal stuff with public OSS project :) 

i dont understand that what you mean by that ... but anyway

that's fine, but what i would expect / want is to have a full featured circuit breaker with monitoring ... what i'm saying is that i would hope THIS toolkit would provide that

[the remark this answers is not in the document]
jakubjanecek: All possible. Just that it first needs to be open source and it shouldn't prevent us from using Monix CircuitBreaker, we can have both.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
tomasherman: well i assume you need F to be :Clock for the circuit breaker ... why inventing this new mechanism for providing the clock when it's already solved - just expose it to the user

jakubjanecek: One of the design decisions of this toolkit is to use constructors for dependencies. I think `Clock` is a regular dependency, I don't see it as a lawful typeclass that should be passed implicitly using context bound. Do we agree on that or not? My view is that if I do this with `Clock` I could as well do it with `Client` (http4s client) - yes, it's possible but I don't like that approach, constructors are much more comprehensible to newcomers.

tomasherman: i generally dislike the pattern of `feel free to give me something and if you don't, i will magic up something for you` ... this might be so simple that it's not an issue but still it rings alarm bells for me

jakubjanecek: I agree but this is exactly the exception I would be willing to make. `Clock` is basically just effectful wrapper over `System.currentTimeMillis` and `System.nanoTime`, so there is nothing to configure or whatever. By making it non-optional everyone will have to create an instance manually and pass it. It just seems unnecessary in this case but if you both think it should be that way I will do it that way.

sideeffffect: my 2 halere: I prefer normal parameter, probably something like this
```scala
def make[F[_]: Sync](clock: Clock[F])(config: CircuitBreakerConfig)
```
but I'm OK with
```scala
def make[F[_]: Sync: Clock](config: CircuitBreakerConfig])
```

jakubjanecek: Sorry but why making it "confusing"? Why should `Clock` be in any way different than any other dependecy? If anything this will be the end result:
```scala
def make[F[_]: Sync](config: CircuitBreakerConfig, clock: Clock[F])
```

sideeffffect: looking at the use of clock `CircuitBreaker.of` expects this as an implicit parameter. In that case, I would honor that style and make the signature like this:
```scala
def make[F[_]: Sync: Clock](config: CircuitBreakerConfig])
```

jakubjanecek: Sorry but I don't want to do that. This library has its own style described in the "design decisions" documentation so I will honor that.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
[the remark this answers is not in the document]
jakubjanecek: 👍

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,13 @@
+package com.avast.sst.monix.catnap
+
+import cats.effect.{Clock, Sync}
+import monix.catnap.CircuitBreaker
+
+object CircuitBreakerModule {
+
+  /** Makes [[monix.catnap.CircuitBreaker]] initialized with the given config and optionally [[cats.effect.Clock]]. */
```
sideeffffect: doc needs to be updated ;)

jakubjanecek: Ohh, thanks.

jakubjanecek:
I've decided to extend this PR after all and add more features. Mainly adding custom CircuitBreaker for http4s Client with logging and metrics (copied from our internal library). This is still not finished, it needs to be tested and documented, but you can have a look if you want to to give me early feedback. Thanks.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,31 @@
...
+                         onOpen)(clock)
+  }
+
+}
+
+object CircuitBreakerModule {
+
+  def apply[F[_]: Sync] = new CircuitBreakerModule[F]
+
+}
```
sideeffffect: why not just 
```scala
object CircuitBreakerModule {
  def make[F[_]: Sync](config: CircuitBreakerConfig,
           clock: Clock[F],
           onRejected: F[Unit] = F.unit,
           onClosed: F[Unit] = F.unit,
           onHalfOpen: F[Unit] = F.unit,
           onOpen: F[Unit] = F.unit): F[CircuitBreaker[F]] = {
    CircuitBreaker[F].of(config.maxFailures,
                         config.resetTimeout,
                         config.exponentialBackoffFactor,
                         config.maxResetTimeout,
                         onRejected,
                         onClosed,
                         onHalfOpen,
                         onOpen)(clock)
  }
}
```
?

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger
+
+import cats.effect.Sync
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics.State
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics.State._
+import io.micrometer.core.instrument.MeterRegistry
+
+class MicrometerCircuitBreakerMetrics[F[_]: Sync](name: String, meterRegistry: MeterRegistry) extends CircuitBreakerMetrics[F] {
```
sideeffffect: shouldn't this class be private?

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
sideeffffect: sorry, I still don't understand. You don't need partial application, because you have just 1 type parameter: `F`, and you _do_ have access to it.

jakubjanecek: I have access to the type parameter but I don't have access to the instance of the type parameter which I need to provide the default values `F.unit`.

jakubjanecek:
I've fixed the build and added some documentation so this could be ready to be merged I think. Please do have a look at the enriched circuit breaker for the http4s `Client` (that's the copied internal implementation) whether it is general enough to be included in this project or whether it needs some tweaking. Thanks.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: I see that you use it in `Http4sClientCircuitBreakerModule`. But I don't you (or any other user) care that it's `MicrometerCircuitBreakerMetrics` specifically, I think they just care that it's an `CircuitBreakerMetrics` instance.

That leads me to think that `MicrometerCircuitBreakerMetrics` being public is an implementation detail leaking out and that the class should be private. You can then create an instance with:
```scala
object MicrometerCircuitBreakerMetrics {
  def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): CircuitBreakerMetrics[F] = new MicrometerCircuitBreakerMetrics(name, meterRegistry)
}
```

http4s-client-monix-catnap-micrometer/src/main/scala/com/avast/sst/http4s/client/monix/catnap/micrometer/Http4sClientCircuitBreakerModule.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  /**  */
+  def make[F[_]: Sync](name: String,
+                       circuitBreakerConfig: CircuitBreakerConfig,
+                       client: Client[F],
+                       meterRegistry: MeterRegistry,
+                       clock: Clock[F]): Resource[F, Client[F]] = {
+    val F = Sync[F]
+
+    class ServerFailure(val response: Response[F], val close: F[Unit]) extends Exception
```
sideeffffect: who is the supposed receiver/consumer of this exception? it seems to me it's just for internal control flow, and thus should also be private.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
sideeffffect: aaaah, now I see :+1:

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,32 @@
+package com.avast.sst.monix.catnap
+
+import cats.effect.{Clock, Sync}
+import monix.catnap.CircuitBreaker
+
+class CircuitBreakerModule[F[_]](implicit F: Sync[F]) {
```
sideeffffect: would this trick also work if the class were private?

http4s-client-monix-catnap-micrometer/src/main/scala/com/avast/sst/http4s/client/monix/catnap/micrometer/Http4sClientCircuitBreakerModule.scala
jakubjanecek: It's inside the method so it is "private".

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: I don't think so, you definitely need some "interface" where you can access the `make` method.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
jakubjanecek: I was thinking of adding the module but I decided not to because most people probably won't use it directly but through the other interop modules. However I did create the module I would leave the class public anyway because you want to return the most specific think in case someone wanted to instantiate it for some specific reason. Again, initialization modules are not an abstraction of any kind and are not really supposed to hide stuff, they are here to simplify the initialization and make it consistent throughout the whole program..

http4s-client-monix-catnap-micrometer/src/main/scala/com/avast/sst/http4s/client/monix/catnap/micrometer/Http4sClientCircuitBreakerModule.scala
sideeffffect: :+1:

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: > you want to return the most specific thing

Quite the opposite. We need to hide implementation details, and expose only "interfaces". Not doing that would be a bad API design. Our users then would have to worry about programming against the API, not implementation. Let's not pass that burden onto them, let's not give them that pseudo-choice to begin with.

(it would also bind our hand with future development of the library)

jakubjanecek: I agree with your opinions generally but this is a different case. I think we've already discussed this once. If I were providing some library feature I would of course provide an interface however this is not a library providing feature but only initialization helper (basically). I want to return the most specific thing because I don't know how the user wants to use that thing. Of course the user will most probably use it "under the interface".

jakubjanecek: We've had long personal discussion about this. I plan to do some changes.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
```diff
@@ -0,0 +1,41 @@
...
+      state match {
+        case Closed   => F.delay(circuitState.set(CircuitClosed))
+        case Open     => F.delay(circuitState.set(CircuitOpened))
+        case HalfOpen => F.delay(circuitState.set(CircuitHalfOpened))
+      }
+    }
+
+  }
+
+}
```
sideeffffect: I think
```scala
object MicrometerCircuitBreakerMetrics {
  private class MicrometerCircuitBreakerMetricsImpl[F[_]: Sync](name: String, meterRegistry: MeterRegistry) extends CircuitBreakerMetrics[F] {
    ...
  }
  def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): CircuitBreakerMetrics[F] = new MicrometerCircuitBreakerMetricsImpl(name, meterRegistry)
}
```
is just enough, there's no need to add more concepts ;) `MicrometerCircuitBreakerMetrics` as defined brings 0 value

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.http4s.client.Client
+import org.slf4j.LoggerFactory
+
+object Http4sClientCircuitBreakerModule {
+
+  private lazy val logger = LoggerFactory.getLogger("Http4sClientCircuitBreakerModule")
+
+  /**  */
+  def make[F[_]: Sync](name: String,
+                       circuitBreakerConfig: CircuitBreakerConfig,
```
sideeffffect: it would probably also be a good idea to pass an actual `CircuitBreaker[F]` instance here (instead of passing config and creating it here on your own)

jakubjanecek: This module is supposed to give you "something more". In this case it's logging and metrics for the circuit breaker therefore I need to instantiate it here. I think there are 2 kinds of modules - base modules and these "interop" modules which give you something on top of just the base modules.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
jakubjanecek: I still think modules should return "specific" thing instead of the most abstract thing. Question: if I remove the now unneeded trait will I be able to introduce it later in backwards-compatible way to allow some specialization of the returned instance? I am not sure... (e.g. let's say I want the returned object to provide the underlying `MeterRegistry`)

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
sideeffffect: I see.
What do you think about splitting creation of the circuit breaker, and wrapping HTTP client with that circuit breaker?

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: > if I remove the now unneeded trait will I be able to introduce it later in backwards-compatible way to allow some specialization of the returned instance

Yep. Change `CircuitBreakerMetrics` in that case.
Or, in a case of a more radical change when changing it would be undesirable, introduce a new interface which inherits from `CircuitBreakerMetrics`, but only if you're in such situation. For example sake, we could name it `CircuitBreakerMetricsWithRegistry`. And then `object MicrometerCircuitBreakerMetrics` would then return an instance of `CircuitBreakerMetricsWithRegistry`.

In general, by returning an instance of an interface, you're also communicating an intent to the users which methods the are supposed to use/rely on, and which not. IMHO these design principles help users to see the intents of the designers without reading long documentation and also potential new contributors to this project.

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
sideeffffect: for the record, I in no way object to introducing a new helper method, which would combine these 2 smaller ones

It's just that I think that it's better to have small, versatile building blocks, and build bigger things from them, compared to having just big unwieldy bricks.

jakubjanecek:
@tomasherman @augi Could you please have a look at this now? Thanks.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: I still think it's worthwhile to have a `make` method to create an instance of `CircuitBreakerMetrics[F]` :wink:

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
sideeffffect: just for curiosity's sake, could you try it? I have small suspicion that it might work, because for that trick, you don't need to create a `CircuitBreakerModule` instance from outside

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
jakubjanecek: I don't understand. I think it's there, see `MicrometerCircuitBreakerMetricsModule`.

docs/monix-catnap.md
```diff
@@ -0,0 +1,18 @@
...
+This module provides interop between Scala Server Toolkit and [monix-catnap](https://monix.io/docs/3x/#monix-catnap) library.
+
+## Circuit Breaker
+
+You can use `CircuitBreakerModule` to instantiate and configure a circuit breaker and you can implement `CircuitBreakerMetrics` to get
+monitoring of the circuit breaker. There is an implementation for Micrometer in `sst-monix-catnap-micrometer` module.
+
+All of this is tied with http4s HTTP client in the `sst-http4s-client-monix-catnap` module so in practice you want to use
+`Http4sClientCircuitBreakerModule` which wraps any `Client[F]` with a `CircuitBreaker` and adds logging/metrics. However
+the most important feature of the enriched circuit breaker is that any HTTP failure (response `5xx`) is converted to an exception internally
```
tomasherman: i think this should be configurable - by that i mean it should be configurable what triggers a failure in circuit breaker - for example `408 Request Timeout` or `429 Too Many Requests` seem like they should trigger CB

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    private val F = Sync[F]
+
+    private[this] val CircuitOpened = -1
+    private[this] val CircuitHalfOpened = 0
+    private[this] val CircuitClosed = 1
+
+    private val accepted = meterRegistry.counter(s"circuit-breaker.$name.accepted")
+    private val rejected = meterRegistry.counter(s"circuit-breaker.$name.rejected")
+    private val circuitState = meterRegistry.gauge[AtomicInteger](s"circuit-breaker.$name.state", new AtomicInteger(CircuitClosed))
```
tomasherman: i think MVar would be more idimoatic for this

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,55 @@
+package com.avast.sst.monix.catnap
+
+import cats.effect.{Clock, Sync}
+import cats.syntax.flatMap._
+import monix.catnap.CircuitBreaker
+import org.slf4j.LoggerFactory
+
+class CircuitBreakerModule[F[_]](implicit F: Sync[F]) {
+
+  private lazy val logger = LoggerFactory.getLogger("CircuitBreaker")
```
tomasherman: why are you providing custom logger name? isn't the standard to use class as logger name?

tomasherman: also, why don't you provide some logger that works well with cats-effect? seems kinda annoying too always having to wrap it yourself

tomasherman: it might also be useful to add `name` into the logger name so that logback can be configured to enable tracing for some problematic upstreams - the name then would be for example `CircuitBreaker.rules-storage`

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Wraps [[org.http4s.client.Client]] with [[monix.catnap.CircuitBreaker]] which provides important metrics and logs breaker events.
+    *
+    * The circuit breaker is special in that it also catches any HTTP 5xx responses and considers them as failures for the circuit breaker.
+    */
+  def make[F[_]: Sync](name: String,
+                       client: Client[F],
+                       circuitBreakerConfig: CircuitBreakerConfig,
+                       circuitBreakerMetrics: CircuitBreakerMetrics[F],
+                       clock: Clock[F]): Resource[F, Client[F]] = {
```
tomasherman: does this really return resource? Seems like this is only wrapping http client so i don't think there is a reason for return `Resource`

docs/monix-catnap.md
sideeffffect: configurable is definitely the way to go. that being said, `5xx` is a good default IMHO

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
sideeffffect: :100:

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: don't you mean `IORef`, or whatever it's called (`MVar`s can be empty, which doesn't make much sense for `circuitState`).
On one hand I would agree, it would be a nice pure FP approach.
On the other hand, it's totally private internal and I would expect `IORef` comes with more overhead than just `AtomicInteger`, and I don't know if it's worth it
:man_shrugging:

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: currently we have:
```scala
// MicrometerCircuitBreakerMetrics.scala
object MicrometerCircuitBreakerMetrics {
  private[micrometer] class Impl[F[_]: Sync](name: String, meterRegistry: MeterRegistry) extends CircuitBreakerMetrics[F] {
    ...
  }
}

// MicrometerCircuitBreakerMetricsModule.scala
object MicrometerCircuitBreakerMetricsModule {
  def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): F[CircuitBreakerMetrics[F]] = {
    ...
  }
}
```

I propose something like this
```scala
// MicrometerCircuitBreakerMetrics.scala
object MicrometerCircuitBreakerMetrics {
  private class MicrometerCircuitBreakerMetricsImpl[F[_]: Sync](name: String, meterRegistry: MeterRegistry, circuitState: IORef[CircuitState]) extends CircuitBreakerMetrics[F] {
    ...
  }
  def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): F[CircuitBreakerMetrics[F]] = {
    for {
      circuitState <- IORef.make(CircuitState.CircuitClosed) // or however you create a new one
      cb = new MicrometerCircuitBreakerMetricsImpl(name, meterRegistry)
    } yield cb
  }
}
```

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
sideeffffect: after a bit of thinking, I would be for the FP approach, the overhead shouldn't be any (or be significant), because it's essentially doing the same thing, and using `IORef` with proper enum communicates the _intent_ better, then the lowlevel mechanics with `AtomicInteger`.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
sideeffffect: > logger that works well with cats-effect

good point! But I must say, that I don't know, if it's a solved problem -- or solved well, to be precise. I know there exist things like `log4cats` or `log-effect`, but I have no idea about their qualities. (I've made a small overview of various logging libs for Scala a while ago, but that's on our internal Scala guild wiki.)

> tracing for some problematic upstreams

:+1:

tomasherman: well at the very least i think you can write a custom wrapper around `Logger`interface so that the users don't have to wrap those things everywhere (might be different PR) ... we also do some support for context propagation during logging but that might be too specific to implement here

docs/monix-catnap.md
tomasherman: sure - another thing to consider is that i think CB should be somehow enabled per IP address and not http client but that's probably a lot more work

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: No, this doesn't work. I tried but I knew it couldn't be done. You get a compilation error `Error:(53, 26) private class CircuitBreakerModule escapes its defining scope as part of type com.avast.sst.monix.catnap.CircuitBreakerModule[F]
  def apply[F[_]: Sync]: CircuitBreakerModule[F] = new CircuitBreakerModule[F]`

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
jakubjanecek: Good point.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
jakubjanecek: Gauge is defined as this:

```java
public <T extends Number> T gauge(String name, T number)
```

It requires `T` to be `Number` and `AtomicInteger` satisfies that. There would be a possibility to use another method and use `Ref` (or whatever) but I don't think it's worth the trouble (I would have to provide a method to convert the reference to `Number` which means I would have to run the effect) considering it is totally internal implementation.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetrics.scala
jakubjanecek: Oh, I see now. The current structure is leftover after some refactorings. Needs to be changed.

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: The "custom" name is a leftover after refactorings. Of course it should be the class name. The reason I use `String` instead of `Class` is that whenever it's part of `object` it looks nasty (dollar sign at the end) so I tried to use the plain version without dollar sign (this does not apply to this case because this is inside `class` but as I said it's some leftover after refactorings).

I don't think the upstream name should be part of the logger name. I do write out the upstream name in the log message though so I think that is sufficient.

And finally the topic about logging library. Of course I was looking for some `F-compatible` logging library but it seems to me there is none that would be mature and spread enough to consider it for this library. Adding a dependency is a burden and I am not ready to do that with a logging library just now. That being sad I only added `slf4j-api` (the lowest common denominator hopefully everyone is ok with) and I use that internally - which is not ergonomic but who cares, it's just a few logging lines. However that does not mean everyone else has to do that. Users of this library can of course choose any logging library they want. This way I am not forcing anyone to use my chosen library but I am also able to log messages inside the library and it's almost dependency-free (`slf4j-api` seems reasonable to me). Makes sense?

tomasherman: > I don't think the upstream name should be part of the logger name. I do write out the upstream name in the log message though so I think that is sufficient.

Having the name part of logger name is strictly better because you can then control the logging using `logback.xml`

about logging library, it would make sense to have a FP-ready slf4j api wrapper readily available to for people but i'm not gonna force you to write one

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics.State
+import com.avast.sst.monix.catnap.CircuitBreakerMetrics.State.{Closed, HalfOpen, Open}
+import io.micrometer.core.instrument.MeterRegistry
+
+object MicrometerCircuitBreakerMetricsModule {
+
+  /** Makes [[com.avast.sst.monix.catnap.CircuitBreakerMetrics]] from [[io.micrometer.core.instrument.MeterRegistry]]. */
+  def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): F[CircuitBreakerMetrics[F]] = {
+    Sync[F].delay(new MicrometerCircuitBreakerMetrics(name, meterRegistry))
```
sideeffffect: why `Sync[F].delay`? is `MicrometerCircuitBreakerMetrics`'s constructor performing side effects?

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    private val F = Sync[F]
+
+    private[this] val CircuitOpened = -1
+    private[this] val CircuitHalfOpened = 0
+    private[this] val CircuitClosed = 1
+
+    private val accepted = meterRegistry.counter(s"circuit-breaker.$name.accepted")
+    private val rejected = meterRegistry.counter(s"circuit-breaker.$name.rejected")
+    private val circuitState = meterRegistry.gauge[AtomicInteger](s"circuit-breaker.$name.state", new AtomicInteger(CircuitClosed))
```
sideeffffect: what do you think about using the `IORef`?

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
tomasherman: btw just something we ran into backend-common, somtimes it pays of to divorce F required to construct something from the F used by the actual thing, eg:

``` scala
  def make[F[_]: Applicative, G[_]: Sync](name: String, meterRegistry: MeterRegistry): F[CircuitBreakerMetrics[G]] = {
```

just something to keep in mind

http4s-client-monix-catnap/src/main/scala/com/avast/sst/http4s/client/monix/catnap/Http4sClientCircuitBreakerModule.scala
```diff
@@ -0,0 +1,51 @@
...
+  def make[F[_]: Sync](name: String,
+                       client: Client[F],
+                       circuitBreakerConfig: CircuitBreakerConfig,
+                       circuitBreakerMetrics: CircuitBreakerMetrics[F],
+                       clock: Clock[F]): F[Client[F]] = {
+    val F = Sync[F]
+
+    class ServerFailure(val response: Response[F], val close: F[Unit]) extends Exception
+
+    val cbResource = CircuitBreakerModule[F].makeLoggingCircuitBreaker(
```
sideeffffect: Why not pass an `CircuitBreaker[F]` as a parameter to `make`, instead of creating it here in place?
(I'm sorry I'm bringing this up again, but I just can't really understand why we can't do that.)

jakubjanecek: It's a bit convoluted because `CircuitBreaker` is not a `trait` and is not straightforward to monitor but I'll try to do something.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
jakubjanecek: This has already been discussed. Gauge requries `Number` so it's not worth the complication to introduce `IORef`.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
jakubjanecek: @sideeffffect To be honest I almost try to wrap everything in `F` just to be sure :) In this case would `new AtomicInteger` count as side effect?

@tomasherman Good point. I did not realize that but I think I'll redo all the modules (if possible) to allow that in some other PR.

docs/monix-catnap.md
jakubjanecek: I'll make it configurable, good point.

I am not sure it's right to do it per IP but even if it was I don't think it's doable in the current implementation. Right now it works on the level of `Request => Response` and the `Request` contains URL with domain (not the actual target IP). We would have to do this inside the HTTP client. We can consider this somehow in the future but I don't want to block this PR with this right now.

jakubjanecek: `408` is actually a client failure (at least according to wiki) so I did not use that but I added `429` as default.

monix-catnap-micrometer/src/main/scala/com/avast/sst/monix/catnap/micrometer/MicrometerCircuitBreakerMetricsModule.scala
sideeffffect: of course you're right @jakubjanecek , `new AtomicInteger(CircuitClosed)` needs to "be behind `Sync[F].delay`", bloody side effects hiding everywhere...

If we wan to make it super obvious, we could do something like this:
```scala
def make[F[_]: Sync](name: String, meterRegistry: MeterRegistry): F[CircuitBreakerMetrics[F]] =
  for {
    counter <- Sync[F].delay(new AtomicInteger(CircuitClosed))
    cbMetrics = new MicrometerCircuitBreakerMetrics(name, meterRegistry, counter)
  } yield cbMetrics

private val CircuitOpened = -1
private val CircuitHalfOpened = 0
private val CircuitClosed = 1 

private class MicrometerCircuitBreakerMetrics[F[_]: Sync](name: String, meterRegistry: MeterRegistry, counter: AtomicInteger) extends CircuitBreakerMetrics[F] {
  ...
}
```
just a suggestion...

jakubjanecek: Yep, it look good.

jakubjanecek:
I think we're ready for another round :)

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
```diff
@@ -0,0 +1,55 @@
...
+    make(config, clock, loggingOnRejected, loggingOnClosed, loggingOnHalfOpen, loggingOnOpen)
+  }
+
+}
+
+object CircuitBreakerModule {
+
+  /** Creates [[com.avast.sst.monix.catnap.CircuitBreakerModule]] specialed for `F[_]: Sync`. */
+  def apply[F[_]: Sync]: CircuitBreakerModule[F] = new CircuitBreakerModule[F]
+
```
sideeffffect: we should add something like
```scala
def withLogging[F[_]: Sync](circuitBreaker: CircuitBreaker[F]): CircuitBreaker[F] = {
  circuitBreaker
    .doOnRejectedTask(F.delay(logger.trace(s"Circuit breaker for $name rejected request.")))
    .doOnClosed(F.delay(logger.trace(s"Circuit breaker for $name closed.")))
    .doOnHalfOpen(F.delay(logger.trace(s"Circuit breaker for $name half-opened.")))
    .doOnOpen(F.delay(logger.trace(s"Circuit breaker for $name opened.")))
}

def withMetrics[F[_]: Sync](circuitBreakerMetrics: CircuitBreakerMetrics[F])(circuitBreaker: CircuitBreaker[F]): CircuitBreaker[F] = {
  circuitBreaker
      .doOnRejectedTask(circuitBreakerMetrics.increaseRejected)
      .doOnClosed(circuitBreakerMetrics.setState(Closed))
      .doOnHalfOpen(circuitBreakerMetrics.setState(HalfOpen))
      .doOnOpen(circuitBreakerMetrics.setState(Open))
}
```
We could then use `withMetrics` in `Http4sClientCircuitBreakerModule.make`, and remove that unnecessary code from `Http4sClientCircuitBreakerModule.makeFromCircuitBreaker`

monix-catnap/src/main/scala/com/avast/sst/monix/catnap/CircuitBreakerModule.scala
jakubjanecek: Well, I don't want to implement yet another logging framework. I believe that one of those libraries will catch up and we'll use it but for now anyone can choose whatever they want and I am not forcing anything on them.

tomasherman: i still think logger name should contain name for the logger

jakubjanecek: It does: https://github.com/avast/scala-server-toolkit/pull/56/files#diff-184610edf024609b740a39c74e1e9b6bR38

PR #71, opened 2019-11-06
feat: Allow configuration of PureConfig naming conventions

jakubjanecek:


jsilva:
looking good

doobie-hikari-pureconfig/src/main/scala/com/avast/sst/doobie/pureconfig/implicits.scala
```diff
@@ -1,3 +1,20 @@
...
+import pureconfig.generic.ProductHint
+
+/** Contains [[pureconfig.ConfigReader]] instances with default "kebab-case" naming convention. */
+object implicits extends ConfigReaders {
+
+  /** Contains [[pureconfig.ConfigReader]] instances with "kebab-case" naming convention.
+    *
+    * This is alias for the default `implicits._` import.
+    */
+  object KebabCase extends ConfigReaders
```
sideeffffect: do we even need the `KebabCase` object? I think `implicits` should be enough.

jakubjanecek: Well, for some reason the bot decided to merge this even though it wasn't approved by you so the discussion is pointless :D Anyway, I thought it would be better to have an alias just to make it consistent. It seemed to me that if I were writing the import and I would let the IDE help me I would be confused to see `CamelCase` and not seeing `KebabCase` and I am not sure whether I would realize that I can just import the `implicits` object. There's documentation so I don't think this harms anyone and it's just better for the users.

PR #104, opened 2019-12-11
feat: Add grpc-server module with PureConfig integration

jakubjanecek:
I have almost no experience with gRPC so this is what I've put together after looking at the documentation and our internal libraries. Starting with the pure gRPC server initialization only - is there anything really important missing?

### DIFF
--- a/grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
+++ b/grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
@@ -1,0 +1,11 @@
+package com.avast.sst.grpc.server
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.Duration
+
+final case class GrpcServerConfig(port: Int,
+                                  handshakeTimeout: Duration,
+                                  maxInboundMessageSize: Int = 4 * 1024 * 1024,
+                                  maxInboundMetadataSize: Int = 8192,
+                                  serverShutdownTimeout: Duration = Duration(10, TimeUnit.SECONDS))
--- a/grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerModule.scala
+++ b/grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerModule.scala
@@ -1,0 +1,43 @@
+package com.avast.sst.grpc.server
+
+import java.util.concurrent.TimeUnit
+
+import cats.effect.{Resource, Sync}
+import io.grpc.{Server, ServerBuilder, ServerInterceptor, ServerServiceDefinition}
+
+import scala.collection.immutable.Seq
+import scala.concurrent.ExecutionContext
+object GrpcServerModule {
+
+  /** Makes [[io.grpc.Server]] (Netty) initialized with the given config, services and interceptors.
+    *
+    * @param services service implementations to be added to the handler registry
+    * @param executionContext executor to be used for the server
+    * @param globalInterceptors that are run for all the services
+    */
+  def make[F[_]: Sync](config: GrpcServerConfig,
+                       services: Seq[ServerServiceDefinition],
+                       executionContext: ExecutionContext,
+                       globalInterceptors: Seq[ServerInterceptor] = List.empty): Resource[F, Server] =
+    Resource.make {
+      Sync[F].delay {
+        val builder = ServerBuilder
+          .forPort(config.port)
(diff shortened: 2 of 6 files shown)


grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.Duration
+
+final case class GrpcServerConfig(port: Int,
+                                  handshakeTimeout: Duration,
+                                  maxInboundMessageSize: Int = 4 * 1024 * 1024,
+                                  maxInboundMetadataSize: Int = 8192,
+                                  serverShutdownTimeout: Duration = Duration(10, TimeUnit.SECONDS))
```
augi: Wouldn't the `java.time.Duration` be more appropriate?

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
```diff
@@ -0,0 +1,11 @@
+package com.avast.sst.grpc.server
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.Duration
+
+final case class GrpcServerConfig(port: Int,
+                                  handshakeTimeout: Duration,
+                                  maxInboundMessageSize: Int = 4 * 1024 * 1024,
+                                  maxInboundMetadataSize: Int = 8192,
```
augi: Are these limits the same as defaults in the underlying library?

augi: Maybe it could be better to make these parameters `Option`al so we wouldn't have to copy these defaults?

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerModule.scala
```diff
@@ -0,0 +1,44 @@
...
+    *
+    * @param services service implementations to be added to the handler registry
+    * @param executionContext executor to be used for the server
+    * @param globalInterceptors that are run for all the services
+    */
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  def make[F[_]: Sync](config: GrpcServerConfig,
+                       services: Seq[ServerServiceDefinition],
+                       executionContext: ExecutionContext,
+                       globalInterceptors: Seq[ServerInterceptor] = List.empty): Resource[F, Server] =
```
augi: Maybe the `global` prefix is not required.

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
jendakol: It was decided the SST will use `scala.concurrent.duration.Duration` instead everywhere.

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
jendakol: `copy these defaults` where? :thinking:

augi: My question is if these defaults have been copied from the underlying library.

jakubjanecek: Yes, they have been copied from the underlying library. I feel this solution is not perfect but I think it's better than to have everything as `Option` so that's why I've decided to do it that away.

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerConfig.scala
jakubjanecek: Both options are possible but we've talked about it with some people and decided for the Scala `Duration`.

In the past I have read some comment that Scala's `Duration` was never supposed to be used as a general time/date data type so I was always preferring `java.time.Duration`. However in this case it's not really "work with time" but only a data type to transfer some duration value correctly and since this is mostly Scala library that will be interoperating with other Scala libraries I suppose choosing Scala's `Duration` is a better option.

grpc-server/src/main/scala/com/avast/sst/grpc/server/GrpcServerModule.scala
jakubjanecek: I wanted to express that it's applied to all the services but you are right that it's not very nice.

PR #108, opened 2019-12-18
fix: PureConfig ProductHint were not picked up correctly

jakubjanecek:


jakubjanecek:
So making `ConfigReaders` a `class`? Is there any real benefit except that you think it's clumsy? For the user the end result is the same.

PR #110, opened 2019-12-19
feat: Use reference config for ssl context

hanny24:


ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
```diff
@@ -14,11 +14,14 @@ import scala.language.higherKinds
 
 object SslContextModule {
 
+  private[this] val referenceConfig = ConfigFactory.defaultReference().getConfig("ssl-config")
```
jakubjanecek: This should be inside `Sync[F].delay` or `lazy`.

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
```diff
@@ -14,11 +14,14 @@ import scala.language.higherKinds
...
+
   /** Initializes [[javax.net.ssl.SSLContext]] from the provided config. */
-  def make[F[_]: Sync](config: Config): F[SSLContext] = Sync[F].delay {
+  def make[F[_]: Sync](config: Config, withReference: Boolean = true): F[SSLContext] = Sync[F].delay {
```
jakubjanecek: Please provide some ScalaDoc explanation of this parameter.

jakubjanecek: And it would be good to use it in the test we have and maybe example (if it is there, I dont't remember) so that people can see it in action.

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
hanny24: That means we need to move it inside `make`. I'm OK with that, I just think it's not worth it.

jakubjanecek: Yes, this will be called only once most probably so let's be pure in this.

hanny24: moved into make

ssl-config/src/main/scala/com/avast/sst/ssl/SslContextModule.scala
hanny24: added tests

PR #119, opened 2020-01-08
[WIP] Improve my Scalafix: RFC

sideeffffect:
This adds many of the Scalazzi rules. (also sorts imports, but that's a small thing.)
What rules do we want to have. Which ones would we like to keep? Which ones would we like to add?
We would need to translate some rules from `scalaz` to `cats`.
Eventually we could migrate away from Wartremover.

Comments and suggestions welcome.

For example, we probably don't want to enforce `noDefaultArgs`, because default args is essential feature of config classes with PureConfig.
On the other hand, we might want to enforce `noValInAbstract`, because it could lead to initialization-related runtime bugs.

EDIT: some of the changes are only of whitespace, so I recommend to turn on _Hide whitespace changes_

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -8,6 +8,7 @@ rules = [
   ExplicitResultTypes
   SortImports
   Disable
+  MissingFinal
 ]
 
 ExplicitResultTypes {
@@ -20,9 +21,71 @@ SortImports.blocks = [
   "scala."
 ]
 
+Disable.symbols = [
+  {
+    regex = "^\\Qscala/collection/mutable\\E.*$"
+    message = "Java collections have better performance, which is what I assume you were trying to do"
+  }
+  {
+    regex = "^\\Qscala/collection/parallel\\E.*$"
+    message = "parallel collections are low performance and block the thread, prefer scalaz.ApplyN"
+  }
+  {
+    regex = "^\\Qscala/math/Big\\E.*$"
+    message = "scala arbitrary precision numbers are broken: https://github.com/scala/bug/issues/9670"
+  }
+  {
+    regex = {
+      includes = [
+        "^\\Qjava/io\\E.*$"
+        "^\\Qscala/io/Source\\E.*$"
+      ]
+    }
+    message = "legacy blocking API, prefer java.nio"
+  }
+  {
+    regex = "^\\Qjava/net/URL#\\E.*$"
+    message = "URL talks to the network for equality, prefer URI"
+  }
+  {
+    regex = {
+      includes = [
+        # overrides not fully implemented: https://github.com/scalacenter/scalafix/pull/634
+        "^.*\\Q#equals().\\E$"
+        "^.*\\Q#hashCode().\\E$"
+        "^.*\\Q#toString().\\E$" # doesn't catch string interpolators...
+        # more to add here...
+      ]
+    }
+    message = "prefer scalaz.{Equal, Show, Liskov, etc}"
+  }
+
+  {
+    symbol = "scala/Enumeration"
+    message = "prefer a sealed abstract class"
+  }
+
+  {
+    regex = {
+      includes = [
+        "^\\Qscala/util/Either.LeftProjection#get().\\E$"
+        "^\\Qscala/util/Either.RightProjection#get().\\E$"
+        "^\\Qscala/util/Try#get().\\E$"
+        "^\\Qscala/Option#get().\\E$"
+        "^\\Qscala/collection/IterableLike#head().\\E$"
+      ]
+    }
+    message = "not a total function"
+  }
+]
+
 Disable.ifSynthetic = [
   "java/io/Serializable"
   "scala/Any"
+  "scala/Product"
+
+  # local type inference + covariant types fires this
+  # "scala/Nothing"
 
   # when upstream broke noImplicitConversion and we don't agree that their
   # implicits are worth the mental burden.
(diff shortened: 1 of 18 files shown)


.scalafix.conf
```diff
@@ -6,16 +6,86 @@ rules = [
...
+  MissingFinal
 ]
 
 ExplicitResultTypes {
   unsafeShortenNames = true
 }
 
+SortImports.blocks = [
+  "java."
+  "*"
```
hanny24: Suggestion: maybe treat com.avast as separate block

project/Dependencies.scala
```diff
@@ -27,6 +27,7 @@ object Dependencies {
   val pureConfig = "com.github.pureconfig" %% "pureconfig" % "0.12.2"
   val scalaTest = "org.scalatest" %% "scalatest" % "3.1.0"
   val scalazzi = "com.github.vovapolu" %% "scaluzzi" % "0.1.3"
+  val sortImports = "com.nequissimus" %% "sort-imports" % "0.3.2"
```
hanny24: Does it work reliably? Last time I tried it it removed some imports that were essential

cassandra-datastax-driver/src/main/scala/com/avast/sst/datastax/DatastaxHelper.scala
```diff
@@ -1,9 +1,9 @@
...
 
-import com.datastax.oss.driver.api.core.config.{DriverOption, ProgrammaticDriverConfigLoaderBuilder => DriverBuilder}
-
-import scala.collection.JavaConverters._
-import scala.concurrent.duration.Duration
+import com.datastax.oss.driver.api.core.config.{DriverOption, ProgrammaticDriverConfigLoaderBuilder => DriverBuilder}
+
+

import scala.collection.JavaConverters._
```
hanny24: ??

.scalafix.conf
```diff
@@ -34,3 +104,127 @@ Disable.ifSynthetic = [
...
+//        }
+//        message = "Deterministic method is not total, must be called via Maybe.attempt, \/.fromTryCatchNonFatal, etc."
+//      }
+//    ]
+//  }
+//]
+
+DisableSyntax {
+  noAsInstanceOf = true
+  noContravariantTypes = true
```
hanny24: This seems too much.

.scalafix.conf
```diff
@@ -34,3 +104,127 @@ Disable.ifSynthetic = [
...
+  noCovariantTypes = true
```
hanny24: This seems too much

.scalafix.conf
```diff
@@ -34,3 +104,127 @@ Disable.ifSynthetic = [
...
+//  noDefaultArgs = true
+  noFinalVal = true
+  noFinalize = true
+  noImplicitConversion = true
+  noImplicitObject = true
+  noIsInstanceOf = true
+  noNulls = true
```
hanny24: It might make integration with java libs very difficult.

.scalafix.conf
```diff
@@ -34,3 +104,127 @@ Disable.ifSynthetic = [
...
+  noReturns = true
+  noSemicolons = true
+  noTabs = true
+  noThrows = true
+  noUniversalEquality = true
+//  noValInAbstract = true
+  noValPatterns = true
```
hanny24: Probably not in tests

.scalafix.conf
```diff
@@ -34,3 +104,127 @@ Disable.ifSynthetic = [
...
+  noVars = true
```
hanny24: Again, seems to harsh.

.scalafix.conf
sideeffffect: :thinking:

project/Dependencies.scala
sideeffffect: has been working reliably for me so far :man_shrugging:

cassandra-datastax-driver/src/main/scala/com/avast/sst/datastax/DatastaxHelper.scala
sideeffffect: problems with LF vs CRLF, I've fixed it already

.scalafix.conf
sideeffffect: do you think it would be _too much_ even if it can be suppressed with an annotation?

sideeffffect:
> disable specific check using comment
yep. with annotation, to be precise. see the diff, I've already did it at some places

.scalafix.conf
hanny24: In that case it's ok

jakubjanecek:
It's a lot of rules at once. I would like to understand what we sign up for. So even though I understand it is a lot of work I would like to see a list of rules that are added with some short explanation why we want to add them (unless it's really obvious). For example the thing with covariant and contravariant types makes no sense to me. Disabling `java.io` also seems very strange - I know it's blocking but blocking does not need to be bad!

jakubjanecek:
Also... what is the relation of import sorting between Scalafmt and Scalafix?

Also, please remove the commented parts - they shouldn't be there and make the PR bigger and more confusing. Thanks.

sideeffffect:
> what is the relation of import sorting between Scalafmt and Scalafix

they complement each other. Scalafmt sorts imports that are on one line (ala `import x.{y, z}`). Scalafix is sorting the import lines.

.scalafix.conf
sideeffffect: unfortunately I don't know to enforce a rule only in `Compile`, and ignore it in `Test` :slightly_frowning_face:

.scalafix.conf
sideeffffect: hmm, I think that in the rare cases when you need `var` and know what you're doing, you can suppress this rule at the place

sideeffffect:
@jakubjanecek I open the PR even with commented rules, so that we can have a discussion about them -- otherwise how could we decide if we want them or not.

I personally don't have a clear idea about what rules to enforce and which would be counterproductive.

Some of the rules are described here: https://github.com/scalacenter/scalafix/blob/master/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntaxConfig.scala#L18

Btw, maybe I should have mentioned it in the beginning -- this scalafix config comes from [`scalazzi`](https://github.com/scalaz/scalazzi/blob/master/scalafix.conf) and in this PR I would like us to discuss it and transform it into something suiting us.
Regarding variance, we all agree, so I've commented it out. Regarding `java.io`, can you please put a comment under the rule?

jakubjanecek:
This PR was split into many smaller PRs which were accepted with some changes.

PR #120, opened 2020-01-10
Improve my Scalafix (1)

sideeffffect:
* SortImports
    * we want all imports to be canonically sorted and we want a tool to do the job, so that we don't have to do that manually
 * better tuned ExplicitResultTypes
    * we want non-private members to always have type annotation -- useful for reading/documentation and for unintended changes if we leave it only for the type inferencer

supersedes parts of https://github.com/avast/scala-server-toolkit/pull/119


.scalafix.conf
```diff
@@ -1,17 +1,21 @@
...
-  DisableSyntax
-  ExplicitResultTypes
-  Disable
+  RemoveUnused
 ]
 
-ExplicitResultTypes {
-  unsafeShortenNames = true
-}
+SortImports.blocks = [
```
sideeffffect: this is how IDEA does it, so I would keep it that way -- we don't want Scalafix and IDEA fight each other

.scalafix.conf
```diff
@@ -34,3 +39,17 @@ Disable.ifSynthetic = [
     message = "not a total function"
   }
 ]
+
+ExplicitResultTypes {
+  unsafeShortenNames = true
```
hanny24: What does it do?

cassandra-datastax-driver-pureconfig/src/main/scala/com/avast/sst/datastax/pureconfig/ConfigReaders.scala
```diff
@@ -1,72 +1,72 @@
...
-  implicit val cassandraDatastaxDriverAdvancedRequestConfigReader: ConfigReader[AdvancedRequestConfig] = deriveReader
-
-  // Profile overridable config
-  implicit val cassandraDatastaxDriverProfileConfigReader: ConfigReader[ProfileConfig] = deriveReader
-  implicit val cassandraDatastaxDriverProfileBasicConfigReader: ConfigReader[ProfileBasicConfig] = deriveReader
-  implicit val cassandraDatastaxDriverProfileAdvancedConfigReader: ConfigReader[ProfileAdvancedConfig] = deriveReader
-  implicit val cassandraDatastaxDriverProfilePreparedStatementsConfigReader: ConfigReader[ProfilePreparedStatementsConfig] = deriveReader
-  implicit val cassandraDatastaxDriverProfileProfileAdvancedRequestConfigReader: ConfigReader[ProfileAdvancedRequestConfig] = deriveReader
-}
+package com.avast.sst.datastax.pureconfig
```
hanny24: Am I blind? What's the change it this file?

jvm/src/test/scala/com/avast/sst/jvm/system/console/ConsoleModuleTest.scala
```diff
@@ -3,9 +3,9 @@ package com.avast.sst.jvm.system.console
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
 
 import cats.effect.SyncIO
+import org.scalatest.funsuite.AnyFunSuite
```
hanny24: I'm not sure I like that scalatest imports are before `scala.*`. I kind of prefer test related imports to be last.

.scalafix.conf
```diff
@@ -1,17 +1,22 @@
...
+  RemoveUnused
 ]
 
-ExplicitResultTypes {
-  unsafeShortenNames = true
-}
+# the same order as IntelliJ does it, we don't want Scalafix and IDEA fight each other
+SortImports.blocks = [
+  "java."
+  "*"
```
jakubjanecek: I wonder why not this order:
`java`
`scala`
*the rest*

That would make much more sense to me.

project/BuildSettings.scala
```diff
@@ -24,7 +24,8 @@ object BuildSettings {
       Wart.DefaultArguments // for constructors for PureConfig
     ),
     ThisBuild / scalafixDependencies ++= Seq(
-      Dependencies.scalazzi // https://github.com/scalaz/scalazzi
+      Dependencies.scalazzi, // https://github.com/scalaz/scalazzi
```
jakubjanecek: The comments should either be next to the dependency definition or removed completely - I don't think they're needed.

jvm/src/main/scala/com/avast/sst/jvm/execution/ConfigurableThreadFactory.scala
```diff
@@ -10,7 +10,9 @@ import com.avast.sst.jvm.execution.ConfigurableThreadFactory.Config
 /** Thread factory (both [[java.util.concurrent.ThreadFactory]] and [[java.util.concurrent.ForkJoinPool.ForkJoinWorkerThreadFactory]])
   * that creates new threads according to the provided [[com.avast.sst.jvm.execution.ConfigurableThreadFactory.Config]].
   */
-class ConfigurableThreadFactory(config: Config) extends ThreadFactory with ForkJoinWorkerThreadFactory {
+class ConfigurableThreadFactory(config: Config, uncaughtExceptionHandler: UncaughtExceptionHandler = LoggingUncaughtExceptionHandler)
```
jakubjanecek: Let's move this change (which is backwards incompatible) to a separate PR. Thanks.

sideeffffect:
some of the changes are only of whitespace, so I recommend to turn on _Hide whitespace changes_
these are changes from CRLF -> LF

.scalafix.conf
sideeffffect: > ``` # the same order as IntelliJ does it, we don't want Scalafix and IDEA fight each other```

jvm/src/test/scala/com/avast/sst/jvm/system/console/ConsoleModuleTest.scala
sideeffffect: I'm afraid `SortImports` can't recognize which imports are test libraries and which are not :man_shrugging:

.scalafix.conf
jakubjanecek: This basically means that whatever IDEA decides to do we will honor it in our rules. There are people using other editors so maybe we should have rules suiting our needs and wishes.

.scalafix.conf
sideeffffect: apparently it's been removed from current version of Scalafix https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html

.scalafix.conf
tomasherman: intellij is the most commonly used by far - so i wouldn't fight it if at all possible ... for once i agree with @sideeffffect :D

that being said i think we are really nitpicking at this point

sideeffffect: I'm sorry for being so stubborn about this -- it's just that I think IDEA and Scalafix not being in sync would impose huge user experience burden onto many people (me included)

build.sbt
```diff
@@ -379,4 +379,4 @@ lazy val sslConfig = project
   )
 
 addCommandAlias("check", "; scalafmtSbtCheck; scalafmtCheckAll; compile:scalafix --check; test:scalafix --check")
-addCommandAlias("fix", "; scalafmtSbt; scalafmtAll; compile:scalafix; test:scalafix; example/mdoc")
+addCommandAlias("fix", "; compile:scalafix ; test:scalafix; scalafmtSbt; scalafmtAll; example/mdoc")
```
jakubjanecek: Is the space `compile:scalafix ;` important or is it just a typo?

sideeffffect: yeah, that's a typo introduced by me flipping the order
fixed :smile:

jakubjanecek:
@tomasherman @hanny24 Anything you want to add or can we merge this?

PR #124, opened 2020-01-20
lint: Introduce scalafix DisableSyntax rules

sideeffffect:
These rules prevent writing problematic Scala code. These's a considerable overlap with WartRemover. Having these rules gives us option to supersede WartRemover with scalafix entirely.

Documentation for the rules is provided at scalafix's website: https://scalacenter.github.io/scalafix/docs/rules/DisableSyntax.html

---

I would like to open discussion about which of the `DisableSyntax` are good for this project. Here's my suggestion, I took inspiration from the [Scalazzi project](https://github.com/scalaz/scalazzi/blob/master/scalafix.conf). Please keep in mind that each rule can be suppressed in-place. I've left some rules commented out for the sake of discussion -- the eventual merge won't contain any commented code :wink:

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -42,10 +42,6 @@ Disable.ifSynthetic = [
 
 DisableSyntax {
   noAsInstanceOf = true
-//  noContravariantTypes = true
-//  noCovariantTypes = true
-//  noDefaultArgs = true
-//  noFinalVal = true // https://github.com/sbt/sbt/issues/1543
   noFinalize = true
   noImplicitConversion = true
   noImplicitObject = true // https://stackoverflow.com/questions/20380800/scala-implicits-resolution-mechanism-is-declaration-order-dependent#comment31809401_20381535
@@ -56,7 +52,6 @@ DisableSyntax {
   noTabs = true
   noThrows = true
   noUniversalEquality = true
-//  noValInAbstract = true  used widely in this library, but should be on in applications
   noValPatterns = true
   noVars = true
   noWhileLoops = true
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -45,7 +45,7 @@ DisableSyntax {
 //  noContravariantTypes = true
 //  noCovariantTypes = true
 //  noDefaultArgs = true
-  noFinalVal = true // https://github.com/sbt/sbt/issues/1543
+//  noFinalVal = true // https://github.com/sbt/sbt/issues/1543
   noFinalize = true
   noImplicitConversion = true
   noImplicitObject = true // https://stackoverflow.com/questions/20380800/scala-implicits-resolution-mechanism-is-declaration-order-dependent#comment31809401_20381535
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -48,7 +48,7 @@ DisableSyntax {
   noFinalVal = true
   noFinalize = true
   noImplicitConversion = true
-  noImplicitObject = true
+  noImplicitObject = true // https://stackoverflow.com/questions/20380800/scala-implicits-resolution-mechanism-is-declaration-order-dependent#comment31809401_20381535
   noIsInstanceOf = true
   noNulls = true
   noReturns = true
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -40,6 +40,29 @@ Disable.ifSynthetic = [
   }
 ]
 
+DisableSyntax {
+  noAsInstanceOf = true
+//  noContravariantTypes = true
+//  noCovariantTypes = true
+//  noDefaultArgs = true
+  noFinalVal = true
+  noFinalize = true
+  noImplicitConversion = true
+  noImplicitObject = true
+  noIsInstanceOf = true
+  noNulls = true
+  noReturns = true
+  noSemicolons = true
+  noTabs = true
+  noThrows = true
+  noUniversalEquality = true
+//  noValInAbstract = true  used widely in this library, but should be on in applications
+  noValPatterns = true
+  noVars = true
+  noWhileLoops = true
+  noXml = true
+}
+
 ExplicitResultTypes {
   fatalWarnings = true
 
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -45,7 +45,7 @@ DisableSyntax {
 //  noContravariantTypes = true
 //  noCovariantTypes = true
 //  noDefaultArgs = true
-  noFinalVal = true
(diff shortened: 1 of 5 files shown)


micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
```diff
@@ -54,7 +54,7 @@ object MicrometerJmxModule {
     override val step: Duration = Duration.ofMillis(c.step.toMillis)
 
     // the method is @Nullable and we don't need to implement it here
-    @SuppressWarnings(Array("org.wartremover.warts.Null"))
+    @SuppressWarnings(Array("org.wartremover.warts.Null", "scalafix:DisableSyntax.null"))
```
jakubjanecek: Why don't we only use Scalafix for this? It's very unfortunate that you have to disable two different rules. Imagine somone not knowing this... you disable WartRemover and it still does not compile so you have to go back and disable also Scalafix. I think every rule should be checked by only single tool.

.scalafix.conf
```diff
@@ -40,6 +40,29 @@ Disable.ifSynthetic = [
   }
 ]
 
+DisableSyntax {
```
jakubjanecek: Unfortunately not all the rules are explained in the documentation :( in which case I am not very happy about having them enabled. For example `noImplicitConversion` is not documented - does that mean that even extension methods via `implicit class` are not allowed? `noFinalVal` is documented but very poorly so I don't understand the reasoning why it shouldn't be allowed.  Why `noImplicitObject`? That seems like something we might want to have in our code. I also don't understand the contravariant and covariant rules... otherwise the rest seems reasonable to me.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
sideeffffect: Is enabling all the desirable features of Scalafix and disabling WartRemover in one PR what you guys are calling for?

Or more gradual approach, with multiple PR, but keeping no overlap what each tool enforces/covers?

jakubjanecek: I don't think it's really important. If you enable null checking in Scalafix you should probably disable it in WartRemover in the same PR.

.scalafix.conf
sideeffffect: > `noImplicitConversion` is not documented - does that mean that even extension methods via implicit class are not allowed

I think I understand where you're coming from, and I haven't tested it, but I don't think that's the case. In Scala community, implicit conversions and implicit classes are considered different things, so I expect scalafix/DisableSyntax to be aligned with this common understanding.
If it turns out that this rule prevents us from creating/using extension methods, I'm all for disabling it.

> `noFinalVal`

I think the [bug report](https://github.com/sbt/zinc/issues/227) explains it -- it's a workaround a bug in Scala compilation toolchain.

> `noImplicitObject`

[This is the explanation that I've found](https://github.com/scalacenter/scalafix/blob/master/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala#L214).  [Here's probably what's an example of this problematic behavior](https://stackoverflow.com/questions/20380800/scala-implicits-resolution-mechanism-is-declaration-order-dependent)

> Implicits without explicit return types are not eligible in implicit searches higher up in the same file. Without this rule, the typechecker runs into cycles very quickly. Implicit object cannot have an explicit type ascribed; my recommendation is to avoid them in favour if implicit vals. – retronym (_a Scala compiler developer_)

jakubjanecek: @sideeffffect Please check that `noImplicitConversion` behaves the expected way.

@hanny24 What do you think about `noFinalVal`? I've never encountered or heard about that issue and it seems strange to me to ban some regular Scala language feature just because incremental compilation might not work properly on your computer. I would remove this rule.

`noImplicitObject` - even though I've also never heard or encountered it I guess we can have this rule to prevent possible unnecessary issues.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
jakubjanecek: Remove WartRemover from here please.

.scalafix.conf
sideeffffect: the Zinc issue is actually explained very very well! The first link in the issue in the documentation https://github.com/sbt/sbt/issues/1543

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
sideeffffect: forgot

.scalafix.conf
jakubjanecek: Yes, that comment explains it well but the impact to the regular Scala developer is quite abstract. I've clicked through the different issues and some of them are closed and the fact that this isn't resolved even after many years indicates that this is not critical issue encoutered often. I am defintely against having it in the ruleset.

sideeffffect: @hanny24 @tomasherman if you have a strong desire to have `noFinalVal` rule enforced, now's the time to speak up :) otherwise I'll be taking it out some time in the afternoon

hanny24: I actually don't know what does `noFinalVal` do...

tomasherman: not really - i don't think we should limit people trying to contribute unless there is really really strong case for it

sideeffffect: aaand... it's gone :smiley:

sideeffffect:
are we happy with the rules now?

.scalafix.conf
jakubjanecek: @sideeffffect From the discussion above I think we're still missing a confirmation that `noImplicitConversion` really bans only what we want and not implicits classes for example.

sideeffffect:
@jakubjanecek the following snippet works fine, I've removed commented rules, I think we're good to merge :rocket: 
```scala
final case class Square(length: Double)
object Square {
  implicit class SquareOps(private val square: Square) extends AnyVal {
    def area: Double = {
      square.length * square.length
    }
  }
  val area: Double = Square(3.5).area
}
```

PR #137, opened 2020-01-31
lint: Introduce scalafix Disable rules for partial (non-total) functions

sideeffffect:
I took inspiration from the [Scalazzi project](https://github.com/scalaz/scalazzi/blob/master/scalafix.conf)

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -80,6 +80,30 @@ Disable.ifSynthetic = [
   }
 ]
 
+Disable.unlessInside = [
+  {
+    safeBlocks = [
+      "cats/syntax/EitherObjectOps.catchNonFatal",
+      "cats/data/Validated.catchNonFatal",
+      "cats/ApplicativeError.catchNonFatal"
+    ]
+    symbols = [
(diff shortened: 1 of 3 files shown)


jakubjanecek:
In practice what does this do? I don't get the connection between `catchNonFatal` and URL encoding/decoding.

sideeffffect:
the static methods (`encode`/`decode`) on these classes are not total, and in some cases throw
that's why calling these should be wrapped in `catchNonFatal`

jakubjanecek:
I don't get why only these two methods are mentioned. There must be many such methods so why just these two were chosen? We are not even using them in our code. Point is that either the list should be "full" or such rule does not make much sense - anyone working with Java APIs will definitely wrap them properly because you never know when exceptions can be thrown.

sideeffffect:
> why only these two methods

because nobody put in the effort to add more. any help/PRs will be heartily appreciated! This is just a start -- my personal plan is to add more functions as time goes by, typically those that are widely used and throw or have any other suspicious behavior.

PR #147, opened 2020-02-11
fix: Remove WartRemover as it was not working properly on Scala 2.13

jakubjanecek:
build: Update dependencies

refactor: Small improvements


jakubjanecek:
After switching to Scala 2.13 the release failed: https://github.com/avast/scala-server-toolkit/runs/436430636?check_suite_focus=true#step:8:763

I spent a lot of time debugging the problems (trying to tweak different SBT settings), trying to find some existing problem in SBT or Scaladoc or something like that related to Scala 2.13 but I was unsuccessful. Finally I found out that the problem lies in WartRemover which even though supports Scala 2.13 it did not work well together with SemanticDB which is necessary for Scalafix. I consider Scalafix better for the future so I have decided to remove WartRemover altogether.

PR #154, opened 2020-02-15
scalafmt edition

sideeffffect:
https://scalameta.org/scalafmt/docs/configuration.html#edition

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,5 +1,5 @@
 version = "2.3.2"
-
+edition = 2020-02
 maxColumn = 140
 assumeStandardLibraryStripMargin = true
 onTestFailure = "To fix this, run 'sbt fixAll' from the project root directory."


jakubjanecek:
Thanks.

I have 2 comments about this:
* Please try to explain why you are making the changes because it might not be clear from the change itself and link to the documentation. I understand you did this because #153 failed.
* I used to have an edition in my Scalafmt configuration but I've removed it very recently because I do not want to stay on an old Scalafmt style. Even though I don't like all the changes coming in newer versions of Scalafmt I think that it is inevitable to change my mind otherwise my Scalafmt style would be more and more out-of-date. Thus I've decided to just keep as little customization as possible and just go with the defaults mostly. So in my opinion it is good that #153 failed because now I can see the problem and I can decide whether I want to just reformat everything or I want to make some tweaks to the configuration (the first option is IMHO preferred).

sideeffffect:
I'm all for having as least custom deviations from Scalafmt defaults as possible (there's only 1 exception that I can think of now)!

On the other hand, I think it's better to track the `edition` (what those defaults are) and Scalafmt version itself independently. That way, upgrading Scalafmt version will only fix bugs or add features, but never "change the default settings" and thus never break build.

jakubjanecek:
As I explained in my previous comment I do not agree and I have exactly opposite view - I want the PR to fail which will make me think about whether to reformat or reconfigure. There is no point in discussing this more so I am closing this ticket.

PR #190, opened 2020-03-25
feat: Implement gRPC server logging and monitoring interceptors

jakubjanecek:


grpc-server-micrometer/src/main/scala/com/avast/sst/grpc/server/micrometer/MonitoringServerInterceptor.scala
```diff
@@ -0,0 +1,59 @@
...
+import java.util.concurrent.{ConcurrentHashMap, TimeUnit}
+
+import io.grpc.ForwardingServerCall.SimpleForwardingServerCall
+import io.grpc._
+import io.micrometer.core.instrument.{MeterRegistry, Timer}
+
+/** Records important gRPC call metrics in [[io.micrometer.core.instrument.MeterRegistry]].
+  *
+  * Metrics:
+  *   - grpc.<method>.current-calls
```
augi: IMHO we must provide _absolute_ metric name in Micrometer (it is not possible to _nest_ metrics names). So I would prefer to have `grpc.<service>.<method>.*`

grpc-server-micrometer/src/main/scala/com/avast/sst/grpc/server/micrometer/MonitoringServerInterceptor.scala
```diff
@@ -0,0 +1,59 @@
...
+  private class CloseServerCall[A, B](prefix: String, start: Long, currentCallsCounter: AtomicLong, delegate: ServerCall[A, B])
+      extends SimpleForwardingServerCall[A, B](delegate) {
+    override def close(status: Status, trailers: Metadata): Unit = {
+      val durationNs = System.nanoTime - start
+      if (status.isOk) {
+        makeTimer(s"$prefix.successes").record(durationNs, TimeUnit.NANOSECONDS)
+      } else {
+        makeTimer(s"$prefix.failures").record(durationNs, TimeUnit.NANOSECONDS)
+      }
+      currentCallsCounter.decrementAndGet()
```
augi: It is just cosmetics...I would prefer to have this call before the `if`, in order to have more precise values.

grpc-server/src/main/scala/com/avast/sst/grpc/server/interceptor/LoggingServerInterceptor.scala
```diff
@@ -0,0 +1,48 @@
...
+      next: ServerCallHandler[ReqT, RespT]
+  ): ServerCall.Listener[ReqT] = {
+    val methodName = call.getMethodDescriptor.getFullMethodName
+    val finalCall = new CloseServerCall(methodName, call)
+    new OnMessageServerCallListener(methodName, next.startCall(finalCall, headers))
+  }
+
+  private class CloseServerCall[A, B](methodName: String, delegate: ServerCall[A, B]) extends SimpleForwardingServerCall[A, B](delegate) {
+    override def close(status: Status, trailers: Metadata): Unit = {
+      if (!status.isOk) {
```
augi: In the original code, we have this:
```java
if (status.getCode() == Status.Code.UNKNOWN || status.getCode() == Status.Code.INTERNAL) {
  logger.error(String.format("Status code of response to be sent to client from method %s not OK. %s: %s", method.getFullMethodName(), status.getCode(), status.getDescription()), status.getCause());
}
else if (!status.isOk()) {
   logger.warn(String.format("Status code of response to be sent to client from method %s not OK. %s: %s", method.getFullMethodName(), status.getCode(), status.getDescription()), status.getCause());
}
else {
   logger.debug("Status code of response to be sent to client from method {} OK: {}", method.getFullMethodName(), status);
}
```
This is because we didn't want to log as `error` when a client sent invalid request (we log it as `warn` instead).

I'm not saying that the logic in the original code is perfect but we could this about these two categories of errors (client vs. server) and log them differently.

jakubjanecek: Oh, I missed that there was different log level.

grpc-server-micrometer/src/main/scala/com/avast/sst/grpc/server/micrometer/MonitoringServerInterceptor.scala
jakubjanecek: According to the docs `getFullMethodName` returns fully qualified name including the service name so we should be fine.

grpc-server-micrometer/src/main/scala/com/avast/sst/grpc/server/micrometer/MonitoringServerInterceptor.scala
```diff
@@ -24,7 +24,7 @@ class MonitoringServerInterceptor(meterRegistry: MeterRegistry) extends ServerIn
       headers: Metadata,
       next: ServerCallHandler[ReqT, RespT]
   ): ServerCall.Listener[ReqT] = {
-    val prefix = s"grpc.${call.getMethodDescriptor.getFullMethodName}"
+    val prefix = s"grpc.${call.getMethodDescriptor.getFullMethodName.replace('/', '-')}"
```
augi: If the only occurrence of `/` as service-method delimiter, maybe it would be better to use `.` instead of `-`, as Micrometer expects `.` as the unified metric separator 🤔

jakubjanecek: I am not sure how many slashes there are in the name so I thought it would be better to replace them with dash and consider that as a single part in the metric name. But both is possible of course.

augi: [According to the code](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/MethodDescriptor.java#L379), the `getFullMethodName` returns `com.avast.mypackage/methodName`. Also [this test](https://github.com/grpc/grpc-java/blob/master/api/src/test/java/io/grpc/MethodDescriptorTest.java#L82) says the same.

So I would still prefer to just replace `/` with `.`. It would also make sense to replace the dots in service name (because it includes package) but it is not required IMHO.

PR #192, opened 2020-03-25
feat: Implement SentryModule and PureConfig integration

jakubjanecek:


http4s-server-blaze/src/main/scala/com/avast/sst/http4s/server/Http4sBlazeServerModule.scala
```diff
@@ -9,6 +9,7 @@ import org.http4s.server.blaze.BlazeServerBuilder
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.Duration
+
```
hanny24: why?

jakubjanecek: There was a missing newline.

sentry-pureconfig/src/main/scala/com/avast/sst/sentry/pureconfig/implicits.scala
```diff
@@ -0,0 +1,20 @@
...
+import pureconfig.generic.ProductHint
+
+/** Contains [[pureconfig.ConfigReader]] instances with default "kebab-case" naming convention. */
+object implicits extends ConfigReaders {
+
+  /** Contains [[pureconfig.ConfigReader]] instances with "kebab-case" naming convention.
+    *
+    * This is alias for the default `implicits._` import.
+    */
+  object KebabCase extends ConfigReaders
```
augi: I don't see any usage of `KebabCase` - is it ok?

jakubjanecek: Yes, this is for the users if they want to use that naming convention, it's not used in SST.

PR #193, opened 2020-03-26
fix: Adjust the Sentry release property format according to the docs

jakubjanecek:


sentry/src/main/scala/com/avast/sst/sentry/SentryModule.scala
```diff
@@ -26,16 +26,20 @@ object SentryModule {
...
     for {
       customizedConfig <- Resource.liftF {
         Sync[F].delay {
           for {
             pkg <- Option(implicitly[ClassTag[Main]].runtimeClass.getPackage)
-            v <- Option(pkg.getImplementationVersion)
-          } yield config.copy(release = Some(v))
+            title <- Option(pkg.getImplementationTitle)
+            version <- Option(pkg.getImplementationVersion)
+          } yield config.copy(release = Some(s"$title@$version"))
```
augi: If I understand this correctly, `release` will not be set if `ImplementationTitle` or `ImplementationVersion` is missing, right?

I have never set `Implementation-Title` - maybe it's my fault. I would consider to fallback to `pkg.getName`.

PR #199, opened 2020-04-01
ci: run missinglinkCheck in CI

sideeffffect:


sideeffffect:
```

**Comment by jakubjanecek** [comment]
I wanted to enable this too but it was quite immature so I waited. We can test it now if it works well...

File: project/BuildSettings.scala [hunk 1]
```diff
@@ -31,6 +32,18 @@ object BuildSettings {
       "-Ywarn-unused", // necessary for Scalafix RemoveUnused rule (not present in sbt-tpolecat for 2.13)
       "-P:silencer:checkUnused"
     ),
+    missinglinkExcludedDependencies ++= List(
```
jakubjanecek, line 35: Why are all these dependencies excluded? Isn't the point of all of this to check the compatibility with everything?

sideeffffect: you can see the detailed logs here
https://gist.github.com/sideeffffect/4c82347ae93053acac6999b6b055a473

they seem to me as essentially false positives -- I guess it's not rare in the Java world to somewhere in your code call to a class/method that might not always be available, typically for logging libraries

jakubjanecek: Alright.

What is the strategy on how to find out whether the report is false positive or not? I am afraid that this might be a problem and lots of false positives would make it too hard to release this project and I don't want that.

sideeffffect: I can't say I'm an expert, I guess we will have to learn as we go.
But from my experience, missinglink has found for me on binary incompatibility that in Circe, which after I suppressed the PFs from Java logging libraries, has stuck out quite obviously.
Probably because it's a Scala thing, which I know has recently released breaking change versoin.

A bad sign would be if we had to start suppressing Scala (let alone TypeLevel) libs. But so far it's only low level Java things and there have never been problems with binary incompatibility with them -- it's always Scala things, mostly TypeLevel jars.

sideeffffect: The fact that there are considerably many FPs is just because sst aggregates sooo many libs. Even a big project (like the one I work on internally) has just a few.

sideeffffect:
btw I finally decided to bite the bullet and start using missinglink, because there's been a new release recently, which I hoped would improve it's practical use (I think it introduced the suppression for FPs -- without the suppression it's not usable at all) and had been bitten by Scala TypeLevel lib binary incompatibility (_again_ 🤬)

also, `sbt-missinglink` is developed by Scala Center (like Scalafix)

jakubjanecek:
I think it's time to merge this ;)

jakubjanecek (approved):

PR #217, opened 2020-04-22
fix: Use the new way of enabling SemanticDB to fix IDEA compile issue

jakubjanecek:


jakubjanecek:
I commited too much. Will create another PR.

PR #244, opened 2020-05-19
fix: Wrong use of Micrometer metrics

jakubjanecek:
feat: New sst-cats-effect subproject with F.time method

I had a discussion with core Micrometer contributor who pointed our wrong use of the library so I am fixing it with this PR.

I also encountered a problem with Scalafix and Scala 2.13 (I don't know why it hasn't happened before) so I am also disabling Scalafix temporarily.

And the last thing is new `sst-cats-effect` subproject which adds one utility method on top of `F` which allows you to measure and record time it took.


sideeffffect:
> problem with Scalafix and Scala 2.13

That's an unfortunate development. Was that a particular rule? Is there an error message from logs somewhere to investigate?
Or is that something that is a known/reported bug?

jakubjanecek:
It's this bug: https://github.com/scalacenter/scalafix/issues/1125

It was encountered not only in SST but also in our internal projects.

cats-effect/src/main/scala/com/avast/sst/catseffect/package.scala
```diff
@@ -0,0 +1,32 @@
...
+import cats.effect.{Bracket, Clock}
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import scala.concurrent.duration.Duration
+
+package object catseffect {
+
+  /**  */
+  implicit class FOps[F[_], A](private val f: F[A]) extends AnyVal {
```
hanny24, line 15: I believe more standard way how to implement this is to move this code into `ClockHelper` (or whatever name), rename this package to include `syntax` name, and reference  `ClockHelper` from here.

jakubjanecek: I am not sure where this particular method belongs so please do check the names and packages I created if it makes sense.

cats-effect/src/main/scala/com/avast/sst/catseffect/syntax/TimeSyntax.scala
```diff
@@ -0,0 +1,38 @@
...
+
+}
+
+object TimeSyntax {
+
+  final class FOps[F[_], A](private val f: F[A]) extends AnyVal {
+
+    /** Measures the time it takes the effect to finish and records it using the provided function. */
+    def time(record: Duration => F[Unit])(implicit F: Bracket[F, Throwable], C: Clock[F]): F[A] = {
+      val unit = TimeUnit.NANOSECONDS
```
jendakol, line 25: What is this for? Why don't you either a) use it directly b) have it as `val` in the `object`?

jakubjanecek: Field definitions are not allowed in value classes. And the only reason to have the variable is to not make a mistake in two places where it's used.

jakubjanecek:
Yes. I've separated it into a standalone function.

PR #246, opened 2020-05-22
feat: Implement Fs2KafkaModule

### LINKED ISSUE #67: Add module for Kafka
We should support initialization of Kafka client.

jakubjanecek:
This PR implements subproject for [FS2 Kafka](https://fd4s.github.io/fs2-kafka/).

Closes #67

### DIFF
--- a/fs2-kafka/src/main/scala/com/avast/sst/fs2kafka/Fs2KafkaModule.scala
+++ b/fs2-kafka/src/main/scala/com/avast/sst/fs2kafka/Fs2KafkaModule.scala
@@ -3,8 +3,6 @@ package com.avast.sst.fs2kafka
 import cats.effect.{Blocker, ConcurrentEffect, ContextShift, Resource, Timer}
 import fs2.kafka._
 
-import scala.util.chaining.scalaUtilChainingOps
-
 object Fs2KafkaModule {
 
   def makeConsumer[F[_]: ConcurrentEffect: ContextShift: Timer, K: Deserializer[F, *], V: Deserializer[F, *]](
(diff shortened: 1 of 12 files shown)


jakubjanecek:
It allows you to "escape" using `ConsumerSettings` and `ProducerSettings` from the underlying `fs2-kafka` client. Of course that means you also have to rewrite the configuration so it's not so easy. But I guess I can add it to the config case class just to be safe.

jakubjanecek:
I'm planning to use it ;) Thanks, I will add a link to the docs.

fs2-kafka/src/main/scala/com/avast/sst/fs2kafka/Fs2KafkaModule.scala
```diff
@@ -0,0 +1,92 @@
+package com.avast.sst.fs2kafka
+
+import cats.effect.{Blocker, ConcurrentEffect, ContextShift, Resource, Timer}
+import fs2.kafka._
+
+object Fs2KafkaModule {
+
+  def makeConsumer[F[_]: ConcurrentEffect: ContextShift: Timer, K: Deserializer[F, *], V: Deserializer[F, *]](
```
jsilva, line 8: this should have an arg for the consumer provider
If I recall it's of type 
```
type ConsumerProvider[F[_]] = Map[String, String] => F[KafkaByteConsumer]
```

used with `.withCreateConsumer(...)` fs2-kafka prop. This is very useful for tests, since one can pass a `MockConsumer`

fs2-kafka/src/main/scala/com/avast/sst/fs2kafka/Fs2KafkaModule.scala
```diff
@@ -0,0 +1,92 @@
...
+      .withMaxPrefetchBatches(config.maxPrefetchBatches)
+      .withPollInterval(config.pollInterval)
+      .withPollTimeout(config.pollTimeout)
+      .withMaxPollInterval(config.maxPollInterval)
+      .withMaxPollRecords(config.maxPollRecords)
+      .withRequestTimeout(config.requestTimeout)
+      .withSessionTimeout(config.sessionTimeout)
+      .pipe(setOpt(blocker)(_.withBlocker(_)))
+      .withProperties(config.properties)
+
```
jsilva, line 45: Miss `.withCreateConsumer`

PR #356, opened 2020-10-20
fix: Default value for maxOrphanRequests in Cassandra driver

jakubjanecek:
I'm getting the following error:

```WARN c.d.o.driver.internal.core.channel.ChannelFactory: [s0] Invalid value for advanced.connection.max-orphan-requests: 24576. It must be lower than advanced.connection.max-requests-per-connection. Defaulting to 256 (1/4 of max-requests) instead.```

I presume it was just a typo originally.


amigarulez (commented):
According to changelog, this value was changed in Datastax 4.5.0. Thanks for fix.

PR #382, opened 2020-11-19
feat: Implement Redis support via Lettuce

### LINKED ISSUE #51: Add module for Redis client
We should have support for some Redis client.

[Lettuce](lettuce.io) seems like a mature option.

jakubjanecek:
Closes #51

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,6 +26,8 @@ lazy val root = project
     jvm,
     jvmMicrometer,
     jvmPureConfig,
+    lettuce,
+    lettucePureConfig,
     micrometer,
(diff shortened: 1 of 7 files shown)


sideeffffect:
Why not just https://github.com/profunktor/redis4cats ?

jakubjanecek:
Because I am using Lettuce. We can have both if we need to, can't we?

sideeffffect:
I'm not stopping you. But clearly `redis4cats` should be recommended over the other and you should consider using it too.
The PR it self is fine.

sideeffffect (approved):

PR #396, opened 2020-11-27
docs(doobie): Warning about autoCommit and transctionIsolation

jakubjanecek:


jakubjanecek:
I've found out this statement is not true. Doobie always sets `autoCommit` to `false`: https://github.com/tpolecat/doobie/blob/f16ffe42df584f4c1b6d1430416d08ad60eacf62/modules/core/src/main/scala/doobie/util/transactor.scala#L70

PR #397, opened 2020-11-27
fix(doobie): remove autocommit config that has no effect

hanny24:


jakubjanecek (approved):
To put this in context. We have discussed this and found out this option has no effect so it's better to have it removed: https://github.com/avast/scala-server-toolkit/pull/396#issuecomment-734732919

PR #424, opened 2021-01-14
fix: Eager initialization of Lettuce resources outside Resource

jakubjanecek:


lettuce/src/main/scala/com/avast/sst/lettuce/LettuceModule.scala
```diff
@@ -14,7 +14,7 @@ object LettuceModule {
 
   /** Makes [[io.lettuce.core.RedisClient]] initialized with the given config and optionally [[io.lettuce.core.resource.ClientResources]]. */
   def makeClient[F[_]: Sync](config: LettuceConfig, clientResources: Option[ClientResources] = None): Resource[F, RedisClient] = {
-    val create = clientResources match {
+    lazy val create = clientResources match {
```
sideeffffect, line 17: You might as well inline the body of `create` where it's used -- it's just one place. Then there could be no danger of side effects happening at the wrong time/place.

sideeffffect (approved):
But this is still OK to merge, if don't want to change it any further.

PR #484, opened 2021-03-08
build: Setup sbt-version-policy

jakubjanecek:


jakubjanecek:
I want to integrate this with #199 so I'm merging this. I'll change anything if you have some comments.

.github/workflows/pull.yml
```diff
@@ -14,4 +14,4 @@ jobs:
         with:
           extraKey: pr-${GITHUB_HEAD_REF}
       - name: Check/Compile/Test
-        run: sbt check
+        run: sbt versionPolicyCheck check
```
sideeffffect, line 17: It would make more sense to put `versionPolicyCheck` into `check`

jakubjanecek: I thought about that but this is supposed to be run on PRs only and we also run `check` during release where you want to run `versionCheck`. So I did it this way.

PR #489, opened 2021-03-11
Setup Sentry correctly for version 4.x

sideeffffect:
Sentry 4.x needs little bit different setup than 1.x
https://docs.sentry.io/platforms/java/migration/#migrating-from-iosentrysentry-1x-to-iosentrysentry-4x


sentry/src/main/scala/com/avast/sst/sentry/SentryConfig.scala
```diff
@@ -1,10 +1,10 @@
 package com.avast.sst.sentry
 
 final case class SentryConfig(
-    dsn: String,
+    dsn: Option[String] = None,
```
jakubjanecek, line 4: Why is this optional now? I don't think it can work without `DSN` so that's why it was mandatory.

sideeffffect: This is allowing the application to rely on the `SENTRY_DSN` environment variable instead of an explicit setting.
The downside of this is that since a programmer is not forced to provide the setting, she _might_ forget to set such a variable, but I was thinking that would happen rarely, if ever.
But no strong feelings either way. What do you think?

jakubjanecek: I would like to keep the original code. Multiple points:
- It's quite strange to have a config with everything optional.
- `SENTRY_DSN` is a fallback behavior of the library and you can as well set it in HOCON.
- This library is supposed to unify the way things are configured so it does not make sense to configure one of the components using env variable only.

jakubjanecek (approved):
Could you also please try to set the `sbt-version-policy` intention so that we can test it works? :)

https://github.com/scalacenter/sbt-version-policy#configure

PR #492, opened 2021-03-14
TimeUtils.timeCase/FOps.timeCase added

jendakol:
Improvement of existing functionality for better error case reporting.

I was thinking about creating a custom sealed trait or using the `ExitCase` directly, but `Either` fits quite well I think.


jakubjanecek (approved):
Thank you for the contribution.

PR #562, opened 2021-05-17
Blocker for micrometer modules

jendakol:


micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
```diff
@@ -13,8 +13,9 @@ import java.time.Duration
 object MicrometerJmxModule {
 
   /** Makes configured [[io.micrometer.jmx.JmxMeterRegistry]]. */
-  def make[F[_]: Sync](
+  def make[F[_]: Sync: ContextShift](
       config: MicrometerJmxConfig,
+      blocker: Blocker,
```
jakubjanecek, line 18: Is it really necessary even for this module?

jakubjanecek (commented):
Thank you for the Scalafix fixes but next time please try to separate these things from the actual PR you want to submit, thanks.

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
jendakol: Since adding the `Blocker` is a backward-incompatible change, I tried to prevent doing it in the future. I think the problem might happen with this module too, even though I have no proof.

jendakol:
The PR was split into two 😉

micrometer-jmx/src/main/scala/com/avast/sst/micrometer/jmx/MicrometerJmxModule.scala
jendakol: Removed 😉

jakubjanecek (approved):

PR #590, opened 2021-06-10
build: Prepare Scala 3 build

jakubjanecek:
Preparation for Scala 3 build.

We are missing stable build of http4s for Scala 3 and we are missing build of PureConfig with automatic derivations (currently the supporrt is only experimental).

There are still some issues in Scala 3 compilation which need to be solved.


jakubjanecek:
@jendakol @sideeffffect I realized that we could probably try to finish this PR. PureConfig already has support for Scala 3 but not for `ConfigWriter` so I decided to remove those few `ConfigWriter

tomasherman:
replaced by #872

PR #594, opened 2021-06-12
fix: Set default value for thread keep-alive to 1 minute

### LINKED ISSUE #593: Consider adding configuration options for blocking executor to ExecutorModule
It would be useful to have the option to configure `coreSize` and `keepAlive` values in `ThreadPoolExecutorConfig` passed to the blocking executor during its creation via ExecutorModule make methods. 
Since by default the `coreSize` is set to 0 and the `keepAlive` to 1000 millis, there might be a large number of blocking threads created and closed if a program schedules repeated task to blocking executor with interval higher than the keepAlive. 

It would be nice to add these parameters to the existing methods with default values set to the current value either directly or wrapped in the `ThreadPoolExecutorConfig`.

jakubjanecek:
feat: Make blocking thread pool executor configurable

Closes #593

### DIFF
--- a/jvm/src/main/scala/com/avast/sst/jvm/execution/ExecutorModule.scala
+++ b/jvm/src/main/scala/com/avast/sst/jvm/execution/ExecutorModule.scala
@@ -53,6 +53,8 @@ class ExecutorModule[F[_]: Sync](
 
 object ExecutorModule {
 
+  private final val DefaultBlockingExecutorConfig = ThreadPoolExecutorConfig(0, Int.MaxValue, allowCoreThreadTimeout = true)
+
   /** Makes [[com.avast.sst.jvm.execution.ExecutorModule]] with default callback executor and extra [[cats.effect.Blocker]] executor
     * for blocking operations.
     */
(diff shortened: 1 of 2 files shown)


jvm/src/main/scala/com/avast/sst/jvm/execution/ExecutorModule.scala
```diff
@@ -53,56 +53,71 @@ class ExecutorModule[F[_]: Sync](
...
 
+  private final val DefaultBlockingExecutorConfig = ThreadPoolExecutorConfig(0, Int.MaxValue, allowCoreThreadTimeout = true)
+
   /** Makes [[com.avast.sst.jvm.execution.ExecutorModule]] with default callback executor and extra [[cats.effect.Blocker]] executor
     * for blocking operations.
     */
   def makeDefault[F[_]: Sync]: Resource[F, ExecutorModule[F]] = {
     for {
       numOfCpus <- Resource.eval(Sync[F].delay(Runtime.getRuntime.availableProcessors))
       coreSize = numOfCpus * 2
```
sideeffffect, line 64: Btw, I think that this is a bit questionable, but that is for a different debate.

I think you've already seen this before :smile_cat: 
https://twitter.com/impurepics/status/987758585722621957?lang=en

sideeffffect (approved):

jvm/src/main/scala/com/avast/sst/jvm/execution/ExecutorModule.scala
jakubjanecek: The theory is that num of CPUs is the right number but in practice (and I've seen this discussed somewhere long in the past) is that a bit higher number might work better. And I think I also got inspired in ZIO runtime. Anyway, I don't think there's much difference so if you show me at least some examples or documents (not just one image ;) ) I think we could change it.

PR #813, opened 2022-01-31
Extend PureConfigModule with overrides from environment variables

nb-ceffa:
This PR wants to provide the possibility to override configuration based on server specifics. For example when you have service distributed in different locations where some of them have a slower connection and you need to change timeouts for them or if you need to override some flags

### DIFF
--- a/pureconfig/src/main/scala/com/avast/sst/pureconfig/PureConfigModule.scala
+++ b/pureconfig/src/main/scala/com/avast/sst/pureconfig/PureConfigModule.scala
@@ -3,33 +3,97 @@ package com.avast.sst.pureconfig
 import cats.data.NonEmptyList
 import cats.effect.Sync
 import cats.syntax.either._
-import pureconfig.error.{ConfigReaderFailure, ConfigReaderFailures, ConvertFailure}
-import pureconfig.{ConfigReader, ConfigSource}
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import com.typesafe.config.{Config, ConfigFactory}
+import pureconfig.error.{
+  ConfigReaderFailure,
+  ConfigReaderFailures,
+  ConvertFailure
+}
+import pureconfig.{
+  CamelCase,
+  ConfigReader,
+  ConfigSource,
+  KebabCase,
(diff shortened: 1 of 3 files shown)


jakubjanecek:
@nb-ceffa Could you please share the context of this change? Why do you need it and what problem are you trying to solve? Thanks.

nb-ceffa:
> @nb-ceffa Could you please share the context of this change? Why do you need it and what problem are you trying to solve? Thanks.

Sorry for the blank description. It's fixed now. 
But in general, we need to configure service based on server specifics because some of them are located far away from the rest of the system. We also have one special server that has more endpoints allowed so we need to be able to change the flag in the config.

jakubjanecek:
OK. Is there a reason why you can't use this behavior? https://github.com/lightbend/config#optional-system-or-env-variable-overrides

Because PureConfig is using Lightbend Config underneath and it already supports overriding with env variables. I prefer to be explicit and specify the env variable names directly in my `reference.conf` file but it also supports "config force" to override anything (as documented in the link).

So I am not missing something I don't think this change is needed.

jakubjanecek:
@nb-ceffa Is your problem solved by my suggestion?

nb-ceffa:
> @nb-ceffa Is your problem solved by my suggestion?

sorry, I had to move my focus to something else. 

I know about the overrides. I wanted to do something similar but with a possibility to set your own environment variable prefix (in case you run multiple services on one server) and override multiple fields by one env variable.

jakubjanecek:
Either I don't understand your use case or you don't need this at all. The reason I am pushing back is that this is quite a lot of custom code that I don't think is needed and I generally don't want SST to contain custom logic, it's mostly just wrapping existing libraries with the same way how to configure them.

Why would you need custom env variable prefix? Each application running on the same server gets its own set of env variables, usually those are not global env variables but variables set just for the process being started so you don't have to worry about conflicts. And you can override multiple fields by the same env variable, just do something like this:

```
field1 = "test"
field1 = ${?MY_VALUE}

field2 = "blabla"
field2 = ${?MY_VALUE}
```

PR #872, opened 2022-04-06
Feat/scala3 hermant

tomasherman:
attempt at scala3 support based on `feat/scala3`.

What is it about:
- we had to change config derivation for scala3 ... it's not source compatible with scala2 so i needed to create specific code for all pureconfig modules for scala-2 and scala-3 codes. See docs here: https://pureconfig.github.io/docs/scala-3-derivation.html

Whats missing:
- [x] rewrite few tests for scala3 that are now just for scala2
- [x] docs about pureconfig and scala3

What had to change:
- in `check` task we now only check `mdoc` task for scala3 (we used to do it for all scala versions) - i don't know how to check it for all considering source code is now diverging for scala2 and scala3
- few scalafix options had to be removed as they are scala2 only. I believe it's ok because scala3 is now the 'main' dialect and therefore we apply scalafix just for it, other scala versions are just xcompiled
- scala3 pureconfig modules no longer support camel case - that is a limitation of pureconfig, we just reflect the change (see https://pureconfig.github.io/docs/scala-3-derivation.html)


tomasherman:
wohoo finally green

project/BuildSettings.scala
```diff
@@ -25,13 +31,13 @@ object BuildSettings {
     description := "Functional programming toolkit for building server applications in Scala.",
     licenses := Seq("MIT" -> url("https://raw.githubusercontent.com/avast/scala-server-toolkit/master/LICENSE")),
     developers := List(Developer("jakubjanecek", "Jakub Janecek", "janecek@avast.com", url("https://www.avast.com"))),
-    scalaVersion := "2.13.8",
-    crossScalaVersions := List(scalaVersion.value, "2.12.15"),
+    scalaVersion := scala3,
```
sideeffffect, line 34: I'm not sure about having Scala 3 as the default is a good idea.
Wouldn't having 2.13 as the default save us a lot of headache with Scalafix (among other things)?

project/BuildSettings.scala
```diff
@@ -25,13 +31,13 @@ object BuildSettings {
...
+    crossScalaVersions := List(scala213, scala212, scala3),
     fork := true,
-    libraryDependencies ++= Seq(
-      compilerPlugin(Dependencies.kindProjector),
+    libraryDependencies ++= (if (!isScala3(scalaVersion.value)) List(compilerPlugin(Dependencies.kindProjector)) else List.empty) ++ List(
       Dependencies.catsEffect,
       Dependencies.scalaCollectionCompat,
+      "org.jetbrains" % "annotations" % "21.0.1", // TODO: this should be compile only dependency!
```
sideeffffect, line 40: TODO :wink:

sideeffffect (approved):
Looks great, thanks for putting in the effort. But aren't we too hard on ourselves with Scala 3 _as the default_?

project/BuildSettings.scala
tomasherman: good catch

project/BuildSettings.scala
tomasherman: idk, in my view scala3 is the future so i would vote for shaping the code towards that ... i would also like to move further towards scala3 code in future PRs using compiler toggles etc

tomasherman:
(marked as draft so that it's not merged until everything is in place)

tomasherman:
ok i think this is ready for review if anyone else is interested

project/BuildSettings.scala
jakubjanecek: My first choice would probably also be `2.13` for now but if everything works and we can compile to both versions the default is probably not that important.

project/Dependencies.scala
```diff
@@ -17,6 +17,7 @@ object Dependencies {
   val http4sDsl = "org.http4s" %% "http4s-dsl" % Versions.http4s
   val http4sServer = "org.http4s" %% "http4s-server" % Versions.http4s
   val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.13.2.2"
+  val jetbriansAnnotations = "org.jetbrains" % "annotations" % "21.0.1"
```
jakubjanecek, line 20:
```suggestion
  val jetbrainsAnnotations = "org.jetbrains" % "annotations" % "21.0.1"
```

jakubjanecek (approved):
Thank you!

project/BuildSettings.scala
tomasherman: it only makes a big difference for mdoc imho ... if its all the same to you, i would try to coarse the code into scala3 compatible code as much as possible so i think scala3 default is the best way to go

tomasherman:
Solved all the comments, im going to merge this once all checks pass. Next steps for me is to massage the code into more scala 3 style of code. I will leave it up to @jakubjanecek to release this once he feels like it

tomasherman:
thanks for taking the time to review

PR #961, opened 2022-08-17
Ember support

karadzhov:


jakubjanecek (approved):
Thank you for your contribution. In general, this is fine, I'd just like to avoid any incompatible changes unless really necessary - so for example the package renames could be avoided. I know it's not perfect but it's better than bugging the users. Could you please update it?

karadzhov:
Actually I didn't rename anything at first but then I thought - wouldn't there be some collision if both clients end up in one project as dependencies? What do you think?

jakubjanecek:
Sure, so let's put the new implementation into separate package but let's keep everything already existing where it was. OK?

PR #982, opened 2022-09-17
Pureconfig Toggle typeclass

nb-ceffa:
This PR adds `Toggle` and `WithConfig`. 

For more details please see https://github.com/avast/scala-server-toolkit/pull/982/files#diff-f197acdd0685d1ec179f803054d18d6cb451bb451d20901c4f5d33c196a71df0R69


jakubjanecek:
Could you please describe what you are trying to achieve with `WithConfig` and `Toggle`? Also the build is failing and it would be nice to add some documentation for this.

pureconfig/README.md
```diff
@@ -0,0 +1,17 @@
+# Toggle
```
jakubjanecek, line 1: This documentation needs to go here https://github.com/avast/scala-server-toolkit/blob/master/site/docs/subprojects/pureconfig.md so that it's part of the generated docs site.

pureconfig/README.md
```diff
@@ -0,0 +1,17 @@
...
+This type class is basically an Option with the possibility to keep the whole configuration in place and change a special boolean flag only. Please check the example below:
```
jakubjanecek, line 2: I don't think `Toggle` is a type class (as in the pattern), it's either a type or a class.

pureconfig/src/main/scala/com/avast/sst/pureconfig/WithConfig.scala
```diff
@@ -0,0 +1,17 @@
+package com.avast.sst.pureconfig
+
+import com.typesafe.config.Config
+import pureconfig.ConfigReader
+
+/** Used to retrieve both parsed configuration object and underlying [[Config]] instance.
+  */
```
jakubjanecek, line 7:
```suggestion
/** Used to retrieve both parsed configuration object and underlying [[Config]] instance. */
```

pureconfig/src/main/scala/com/avast/sst/pureconfig/WithConfig.scala
```diff
@@ -0,0 +1,17 @@
...
+final case class WithConfig[T](value: T, config: Config)
```
jakubjanecek, line 8: Please document also usage of this class, thanks.

pureconfig/src/test/scala-2/com/avast/sst/pureconfig/ToggleTest.scala
```diff
@@ -0,0 +1,112 @@
+package com.avast.sst.pureconfig
+
+import cats.{Applicative, FlatMap, Functor, Monad, Monoid}
+import com.avast.sst.pureconfig.util.Toggle
+import com.avast.sst.pureconfig.util.Toggle.{Disabled, Enabled}
+import org.scalatest.diagrams.Diagrams
+import org.scalatest.funsuite.AnyFunSuite
+
+@SuppressWarnings(Array("all"))
```
jakubjanecek, line 9: Why do all warnings have to be suppressed here? 🤔

pureconfig/src/test/scala-3/com/avast/sst/pureconfig/ToggleTest.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import cats.{Applicative, FlatMap, Functor, Monad, Monoid}
+import com.avast.sst.pureconfig.util.Toggle
+import com.avast.sst.pureconfig.util.Toggle.{Disabled, Enabled}
+import com.avast.sst.pureconfig.util.Toggle.TogglePureConfigInstances.*
+import com.typesafe.config.ConfigRenderOptions
+import org.scalatest.diagrams.Diagrams
+import org.scalatest.funsuite.AnyFunSuite
+
+@SuppressWarnings(Array("all"))
```
jakubjanecek, line 11: And here?

pureconfig/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```
+case class Config(myService: Toggle[ServiceConfig])
+case class ServiceConfig(url: String, timeout: Duration)
+```
+
+```
+my_service {
```
jakubjanecek, line 10: Isn't the default casing like this?

```suggestion
my-service {
```

pureconfig/README.md
jakubjanecek: Also this whole sentence does not seem very clear. The example makes it clear but please try to describe why this exists and how it can be used (I think the relation to `Option` might not be important too much).

jakubjanecek (changes requested):
Thank you for your contribution!

I do have some requests for improvement. Please, try to have another look at it.

pureconfig/README.md
nb-ceffa: Please let me know whether this one is more understandable or not:
https://github.com/avast/scala-server-toolkit/pull/982/files#diff-f197acdd0685d1ec179f803054d18d6cb451bb451d20901c4f5d33c196a71df0R69

pureconfig/src/main/scala/com/avast/sst/pureconfig/WithConfig.scala
nb-ceffa: Is this ok? https://github.com/avast/scala-server-toolkit/pull/982/files#diff-f197acdd0685d1ec179f803054d18d6cb451bb451d20901c4f5d33c196a71df0R93

pureconfig/README.md
jakubjanecek: Much better! Thanks.

site/docs/subprojects/pureconfig.md
```diff
@@ -64,3 +64,46 @@ implicit val serverConfigurationReader: ConfigReader[ServerConfiguration] = Conf
 
 val maybeConfiguration = PureConfigModule.make[Task, ServerConfiguration]
 ```
+
+
+## Toggle
+Sometimes is useful to be able to disable some functionality via config. That usually means to add a new boolean field to particular configuration case class. You have to be aware about this new field when you work with the configuration and check the value of it to enable/disable particular functionality.
```
jakubjanecek, line 70:
```suggestion
Sometimes it is useful to be able to disable some functionality via config. That usually means to add a new boolean field to particular configuration case class. You have to be aware of this new field when you work with the configuration and check the value of it to enable/disable particular functionality.
```

site/docs/subprojects/pureconfig.md
```diff
@@ -64,3 +64,46 @@ implicit val serverConfigurationReader: ConfigReader[ServerConfiguration] = Conf
...
+
+`Toggle` is more safe approach for this pattern as it can be either `Enabled[T]` or `Disabled`.
```
jakubjanecek, line 72:
```suggestion
`Toggle` is a more safe approach for this pattern as it can be either `Enabled[T]` or `Disabled`.
```

site/docs/subprojects/pureconfig.md
```diff
@@ -64,3 +64,46 @@ implicit val serverConfigurationReader: ConfigReader[ServerConfiguration] = Conf
...
+So it will force you to deal with the possibility of disabled functionality.
+It also simplifies the workflow and particular config case class as it does not require any changes on it.
```
jakubjanecek, line 74:
```suggestion
It also simplifies the workflow and particular config case class as it does not require any changes to it.
```

site/docs/subprojects/pureconfig.md
```diff
@@ -64,3 +64,46 @@ implicit val serverConfigurationReader: ConfigReader[ServerConfiguration] = Conf
...
+my-service {
+    enabled = false
+    url = "http://www.example.com/v1/some-endpoint"
+    timeout = 5s
+}
+```
+
+
+## WithConfig
+When you need to deal with legacy code, it sometimes happens that you need to load configuration as a standard typesafe's `Config` object and also as your custom object. Reading config with `WithConfig` case class ensures to load of both objects simultaneously.
```
jakubjanecek, line 94:
```suggestion
Sometimes you need to load the configuration both as your custom configuration class and raw `Config` when you deal with legacy code. Using `WithConfig` case class allows you to load both objects simultaneously.
```

jakubjanecek (approved):
Thank you very much. It's ok now. I just have some suggestions in the documentation.

